PR #3, opened 2022-01-11
Use `content-visibility` to contain expensive paints

natemoo-re:
Attempting to see if `content-visibility: auto` help improve scroll performance


natemoo-re:
Well it certainly doesn't _hurt_ performance! Going to merge this.

PR #10, opened 2022-01-17
wire up and style careers to pull role data from markdown files

rdkulp:
This PR used the same approach as blog posts to pull in data from markdown files to add open roles and provide a salary range. As the company grows and more roles open, it would make sense to use the department that gets passed in to organize the roles by department on the careers page.

Careers Page
Screen Shot 2022-01-17 at 11 21 35 AM

Role Page
Screen Shot 2022-01-17 at 11 19 17 AM


rdkulp:
I saw that the deploy failed with the following error for npm run build:
"Invalid time value
RangeError: Invalid time value"

I ran it locally and received the same error and am looking back over the changes I made to see where it could have gone sideways.

PR #13, opened 2022-01-20
adds morphdom

boehs:
Screenshot from 2022-01-20 15-07-07


jonathantneal:
This is only for dev? I’ll let @natemoo-re make the call.

natemoo-re:
Yeah this seems like a problem with `astro` missing the dependency?

PR #16, opened 2022-01-21
Improve analytics script greatly

boehs:
this change is important because tracking people is a total *jam*


matthewp:
Hah, this is fun. I would worry that it violates copyright though :/

boehs:
I'm not well versed in law but would de minims apply? Perhaps it is excessive to have the whole song, what if it was shortened to 10 seconds?

okikio:
Hotel Trivago, 
image

FredKSchott:
This is really funny :) I think we're going to avoid playing an mp3 for now, but thanks for the contribution :D

PR #17, opened 2022-01-21
add repl link to the new websites navbar

okikio:
add repl link to the new websites navbar

### DIFF
--- a/src/components/Nav.astro
+++ b/src/components/Nav.astro
@@ -6,7 +6,8 @@ const { invert = false, marginBottom = false } = Astro.props;
 
 const items = [
     { href: '/blog', title: 'Blog' },
-    { href: 'https://docs.astro.build', title: 'Docs' }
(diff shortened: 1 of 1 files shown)


jonathantneal (approved):
I want @natemoo-re to glance, but +1 from me.

Princesseuh:
This might cause an issue on mobile devices as the menu is currently only designed for a small amount of elements

okikio:
@princesseuh I've updated the name of the link from "Playground" to "Repl", what do you think?

Princesseuh:
I suspect that it still might be an issue on very thin phones but, there's no real solutions that isn't "implement a mobile menu" so 🤷‍♀️

I think that Playground look nicer and is more in line with other tools

okikio:
I'll probably need to contact nate to confirm how he'd like us to approach this, without wrecking the design

FredKSchott:
fixed the CSS to fit on mobile! +1 from me

PR #20, opened 2022-01-25
Blog post on the new static build

matthewp:
This post is meant to:

- Tell people in the wider Astro community about the static build and the benefits.
- Let them know that this is going to the main build in the future.
- Encourage them to report bugs so we can stabilize it enough to make it the main build.

### DIFF
--- a/src/posts/experimental-static-build.md
+++ b/src/posts/experimental-static-build.md
@@ -7,7 +7,7 @@ authors:
 lang: 'en'
 ---
 
-Astro's about to get a lot faster with a newly built build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recently releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
+Astro's about to get a lot faster with a new build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recently releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
 
 - Improves build times by up to 75%.
 - Sets us up for server-side rendering (SSR) in the near future.
--- a/src/posts/experimental-static-build.md
+++ b/src/posts/experimental-static-build.md
@@ -20,4 +20,4 @@ We're __now__ at the stage where this feature is ready for wider use. If you are
 astro build --experimental-static-build
 ```
 
-This build will remain flagged for the next few releases until we iron out all of the kinks. Please help us by reporting issues you encounter, either in the [Discord](https://astro.build/chat) or by filing an [issue](https://github.com/withastro/astro/issues/new/choose).
+This build approach will remain flagged for the next few releases until we iron out any issues, at which point we plan to promote it to be the default `astro build` command. Please help us by reporting issues you encounter, either in the [Discord](https://astro.build/chat) or by filing an [issue](https://github.com/withastro/astro/issues/new/choose).
--- a/src/posts/experimental-static-build.md
+++ b/src/posts/experimental-static-build.md
@@ -1,0 +1,23 @@
+---
+title: 'Scaling Astro Sites to Thousands of Pages'
+description: 'A new experimental flag in Astro build enables building sites with tens of thousands of pages.'
+publishDate: 'January 25, 2022'
+authors:
+  - matthew
+lang: 'en'
+---
+
+Astro's about to get a lot faster with a newly built build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recently releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
+
+- Improves build times by up to 75%.
+- Sets us up for server-side rendering (SSR) in the near future.
+
+This new build works by first building an SSR version of your app and then rendering each page to HTML. Because the site is pre-optimized it can render each page in parallel and will never run out of memory.
+
+We're __now__ at the stage where this feature is ready for wider use. If you are a current Astro user please try out this new build by passing this flag in your build script:
+
+```shell
+astro build --experimental-static-build
+```
+
+This build will remain flagged for the next few releases until we iron out all of the kinks. Please help us by reporting issues you encounter, either in the [Discord](https://astro.build/chat) or by filing an [issue](https://github.com/withastro/astro/issues/new/choose).
--- a/src/posts/experimental-static-build.md
+++ b/src/posts/experimental-static-build.md
@@ -7,7 +7,7 @@ authors:
 lang: 'en'
 ---
 
-Astro's about to get a lot faster with a new build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recently releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
+Astro's about to get a lot faster with a new build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recent releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
 
 - Improves build times by up to 75%.
 - Sets Astro up to support server-side rendering (SSR) in the near future.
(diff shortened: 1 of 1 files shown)


src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
+---
+title: 'Scaling Astro Sites to Thousands of Pages'
+description: 'A new experimental flag in Astro build enables building sites with tens of thousands of pages.'
+publishDate: 'January 25, 2022'
+authors:
+  - matthew
+lang: 'en'
+---
+
+Astro's about to get a lot faster with a newly built build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recently releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
```
sarah11918, line 10: " ... pay attention to RECENT releases..."

jasikpark, line 10:
```suggestion
Astro's about to get a lot faster with a new build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recently releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
```

brevity? built build sounds weird to me :p

jasikpark (commented):
It'd be cool to have some light demos for it.. to show off the SPEEEEED 🚀

src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
+---
+title: 'Scaling Astro Sites to Thousands of Pages'
```
natemoo-re, line 2:
```suggestion
title: 'Scaling Astro's Build Process'
```

Slightly tighter title, totally just IMO.

src/posts/experimental-static-build.md
natemoo-re: Nit: `Astro's` => `Astro is`

src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
...
+description: 'A new experimental flag in Astro build enables building sites with tens of thousands of pages.'
+publishDate: 'January 25, 2022'
+authors:
+  - matthew
+lang: 'en'
+---
+
+Astro's about to get a lot faster with a newly built build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recently releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
+
+- Improves build times by up to 75%.
```
natemoo-re, line 12: What's the real world impact? I'd love something like "... up to 75%. A build that previously took 3 minutes should now take about 45 seconds." Stats based on a 75% reduction, but you might have real numbers.

src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
...
+- Sets us up for server-side rendering (SSR) in the near future.
```
natemoo-re, line 13:
```suggestion
- Sets Astro up to support server-side rendering (SSR) in the near future.
```

src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
...
+
+This new build works by first building an SSR version of your app and then rendering each page to HTML. Because the site is pre-optimized it can render each page in parallel and will never run out of memory.
+
+We're __now__ at the stage where this feature is ready for wider use. If you are a current Astro user please try out this new build by passing this flag in your build script:
+
+```shell
+astro build --experimental-static-build
+```
+
+This build will remain flagged for the next few releases until we iron out all of the kinks. Please help us by reporting issues you encounter, either in the [Discord](https://astro.build/chat) or by filing an [issue](https://github.com/withastro/astro/issues/new/choose).
```
natemoo-re, line 23:
```suggestion
This build approach will remain flagged for the next few releases until we iron out any issues, at which point we plan to promote it to be the default `astro build` command. Please help us by reporting issues you encounter, either in the [Discord](https://astro.build/chat) or by filing an [issue](https://github.com/withastro/astro/issues/new/choose).
```

natemoo-re (commented):
Looks great! A few small suggestions to tighten things up, not blocking.

src/posts/experimental-static-build.md
matthewp: It depends on usage. If you are doing something like reading from an API (Pokemon demo) to generate pages you will get 75%. If you have lots of markdown files in your project you might get a smaller improvement. And if you have a really small project (like only a few pages) it might even be slower (2.5 seconds instead of 2 seconds, for example). So we can't commit to always having a certain build improvement, it depends on the project.

src/posts/experimental-static-build.md
FredKSchott: I like the idea of a number (numbers always hit better). If we want more concise, maybe: 

`title: 'Scaling Astro to 10,000+ Pages'`

src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
+---
+title: 'Scaling Astro Sites to Thousands of Pages'
+description: 'A new experimental flag in Astro build enables building sites with tens of thousands of pages.'
+publishDate: 'January 25, 2022'
+authors:
+  - matthew
+lang: 'en'
+---
+
+Astro's about to get a lot faster with a new build system intended to scale to tens, or even hundreds, of thousands of pages. If you hang out in our [Discord](https://astro.build/chat) or pay attention to recent releases you might have seen a lot of discussion about a "static build". This is our new implementation of `astro build` that does 2 things:
```
FredKSchott, line 10: For a short, informational post like this it's good to get to the point at the start, in 1-2 sentences. Next.js does a good job of this in their blog posts, for example: https://nextjs.org/blog/next-12

Something like:

```
Astro is about to get a lot faster! Our new build optimization process is ready to try out in Astro today:

'''
astro build --experimental-static-build
'''

What's new? ... (this is where you can get into the prose that you have in the first paragraph).
```

src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Improves build times by up to 75%.
+- Sets Astro up to support server-side rendering (SSR) in the near future.
```
FredKSchott, line 13: I'm not sure if most users care about this! Maybe as a point in the paragraph below (since you talk about server-side rendering there) but the main reason that a user would care about this is the speed and scaling improvements.

I also wouldn't shy away from comparing this to how Astro currently works. It can really hit home why this is a big deal, vs. "huh, I guess I thought that was how Astro already works?"

Something like:

```
Previously, Astro would build every page to HTML and then optimize each of those pages. This worked fine for small sites, but became a problem as sites grew larger.
```

src/posts/experimental-static-build.md
```diff
@@ -0,0 +1,23 @@
...
+
+This new build works by first building an SSR version of your app and then rendering each page to HTML. Because the site is pre-optimized it can render each page in parallel and will never run out of memory.
+
+We're __now__ at the stage where this feature is ready for wider use. If you are a current Astro user please try out this new build by passing this flag in your build script:
```
FredKSchott, line 17: nit: This paragraph would read better without `We're __now__ at the stage where this feature is ready for wider use.` and bringing the next sentence front-and-center.

src/posts/experimental-static-build.md
matthewp: Hm, I'm not sure I agree (or maybe I don't understand). You're saying that users don't care about this setting Astro up for SSR but they *do* care about internal implementation details of the build? I sort of see it the opposite. I can see a technical article that explains the *how* of it being faster, but I wouldn't think that a typical dev cares about that. But I would expect them to care about SSR.

FredKSchott: Fair enough! You can make the final call here. It's a trap that I always fall into so I'm extra cautious: does an external audience care about the big, long-term strategic reason for the improvement vs. the improvement itself (speed & scaling)? I think it's great to mention somewhere, but my comment was mainly directed at the fact that it's a high-level callout here.

FredKSchott: Also now that i re-read this, is "improve build times by up to 75%" a good reflection of the improvement being made here? The title mentions scaling, but the call out here is overall speed. Maybe rewrite to better connect to the point made in the title? Something like: "Improves build times, especially for large site"

matthewp: I switched this point to be able lowering memory usage instead since that aligns with the immediate value.

PR #21, opened 2022-02-02
update careers page, about page, style cleanup

FredKSchott:
- adds an "about" page (still a bit short, but can be grown over time)
- replaces our "careers" page with a more general "company" page
- adds a list of contributors to our about page (woo!)
- lots of small cleanup, including better styles for non-homepage pages

Screen Shot 2022-02-01 at 10 00 59 PM


zadeviggers:
I was looking at it (it looks very cool) and noticed that the contributor's name didn't come up in a title-tooltip when you hover on their avatar, which seems like a pretty crucial feature.

FredKSchott:
Good idea, added! Also, avatars are now clickable to go to GitHub URL

PR #26, opened 2022-02-17
Disabling the hero background grid animation for Firefox

tony-sull:
Firefox really chokes on the background grid animation, possibly related to [this](https://bugzilla.mozilla.org/show_bug.cgi?id=739176) older Gecko issue

Let's just disable the animation in Firefox, I'd hate to change the design here - the transforms and blur filters don't seem to be a problem without the animation

### DIFF
--- a/src/components/landing/Grid.astro
+++ b/src/components/landing/Grid.astro
@@ -56,6 +56,12 @@ const { class: className } = Astro.props;
     animation: move var(--time) infinite linear;
 }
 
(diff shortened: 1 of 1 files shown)


jonathantneal:
Nit, but this page crashes my Safari browser. How would we feel about an overall pause?

jonathantneal (approved):

natemoo-re (approved):
Works for me! Glad you're working on this.

PR #31, opened 2022-02-19
Update astro-023.md

leoj3n:
Clarify the dangerous vs trusted HTML example code.

Precedent is already set with that line: `<div set:html={trustedOrAlreadyEscapedHtml}></div>`.

This just expands that to the rest of the example.


natemoo-re (approved):
Definitely in favor of this clarification! Thanks for the PR!

PR #33, opened 2022-02-22
General cleanup and prep for TailwindCSS

tony-sull:
⚠️ This ended up being a bit of a large PR based on files touched, sorry!

- moved component-specific styles out of `main.css`
- removed Prism styles now that we're using Shiki
- refactored page layouts (a bit of duplicate code cleanup, should help when adding Tailwind)
- moved all scripts into `src`, refactored them a bit, and moved component-specific scripts out of the base layout


src/components/PixelLink.astro
```diff
@@ -6,6 +6,13 @@ const { variant = 'primary' } = Astro.props;
 const { before, after } = Astro.slots;
 ---
 
+<!-- Blocking script -->
+<script global hoist>
```
natemoo-re, line 10: Does `global` do anything?

natemoo-re (approved):
This looks amazing, great work!

src/components/PixelLink.astro
tony-sull: Great catch, it doesn't!  I was trying out a bunch of different script setups and didn't notice this had both flags :+1:

FredKSchott:
Hate to do this but there were some bugs in the blog that broke some posts, including the v0.23 release post. I just reverted the commit, if you can resolve the issues in #35 then I think this PR can be re-merged!

PR #41, opened 2022-03-03
Add SPA mode

natemoo-re:
Testing [`micromorph`](https://github.com/natemoo-re/micromorph) for a SPA mode


natemoo-re:
Meh. Not super happy with how this works. Something about the way we're bundling styles makes this stutter on navigation, which is not good.

Would be nice for https://github.com/withastro/astro.build/pull/40 but I wouldn't use it until the stuttering is ironed out.

PR #52, opened 2022-03-25
migrating blog posts to use @tailwindcss/typography

tony-sull:
This updates the blog post pages to use tailwind, including the `@tailwindcss/typography` as a starting point with styles modified to better match our existing designs


FredKSchott:
nit: the page width is wider than on our main site, and a bit hard for my eyes to scan. Revert to current width?

PR #73, opened 2022-04-07
WIP: optimizing images with astro-imagetools

tony-sull:


tony-sull:
Will revisit!  Blog post images are automatically optimized with astro-imagetools, will need to clean those up a bit first

PR #74, opened 2022-04-07
Post: Community Day

tony-sull:


src/pages/blog/community-day.md
```diff
@@ -0,0 +1,80 @@
+---
+layout: '../../layouts/Post.astro'
+title: Community Day
+description: 'After a long week of new features and releases, we wanted to  turn our attention back to the most important aspect of this project: the Astro community.'
```
tony-sull, line 4: @FredKSchott I pulled this from the first paragraph but you may want to update it

src/pages/blog/community-day.md
```diff
@@ -0,0 +1,80 @@
...
+- **sarah11918,** for leading the charge on our docs site revamp, joining aFuzzyBear for the “[This Week in Astro](https://www.youtube.com/playlist?list=PLowBAHWxr_2BsoRUzl5uFwQk7-qWxZGts)” video series, and countless other contributions across GitHub and Discord.
+- **ZV**, for significant contributions in both support and documentation.
+
+We’ll be posting this same list in our Discord later today, along with more details and instructions on how to claim your award if you’re one of the people listed here. Congratulations to all of our winners and nominees! 🥳
+
+## Supporting our Essential Dependencies
+
+Some open source projects are lucky, and can raise large amounts of money from sponsors. [Webpack raises $223,113/yr.](https://opencollective.com/webpack) [Babel raises almost $300,000/yr](https://opencollective.com/babel). 
+
+Unfortunately, these types of projects are in a very small minority. For a variety of reasons, most other funded projects receive only a small fraction of this kind of money. 80% of npm’s top 1000 packages received zero funding of any kind (2019).
```
tony-sull, line 51: @FredKSchott I removed the footnote here, do we want to link an original source?

src/pages/blog/community-day.md
FredKSchott: Oh good catch, I'd use the Twitter catchy first tweet instead:

description: 'Happy Astro Contributor Day! Who are some of our awesome contributors? Where does Astro's sponsorship 💰 go? How does Astro give back to OSS? Lets get into it!'

also: title should be "Astro Contributor Day" or "Happy Astro Contributor Day!", whichever you think is best.

src/pages/blog/community-day.md
FredKSchott: The problem is I don't really have the original data used to calculate it, this was from a webpage I built back in 2019. it could probably be recreated but I figured the footnote was enough of an answer to not be a problem, even if it wasn't entirely satisfying. Thoughts?

tony-sull: Yep I think that's fine!  I removed the footnote thinking it was temporary but I'll add that back in :+1:

PR #81, opened 2022-04-08
Remove `noreferrer`

MarcusOtter:
As discussed on Discord [in #general](https://ptb.discord.com/channels/830184174198718474/830184175176122389/962105868516741163) it would be nice to give Astro credit where credit is due and show when traffic comes from Astro! Right now traffic to the pages in the showcase list looks like "Direct traffic" in analytics tools.

I looked around and couldn't find any implications for this, but I'm happy to reconsider if this is here for a reason.

Fun fact: we could also technically remove `noopener` since all links with `target="_blank"` automatically gets `rel="noopener"`, but it's probably best to keep it in case we ever want to remove `target="_blank"`.


MarcusOtter:
Can confirm that I am now getting the referrer in my analytics:
bild

tony-sull:
Great catch!  I've hung onto `noopener` out of habit, looks like in the push to get everything ready for launch week I defaulted to the older habit of always including `noreferrer` as well 😊

PR #115, opened 2022-05-06
new nav layout

FredKSchott:
- I keep not knowing which links live where, so I took a pass at the header links on the site. 
- I think this makes more sense!

### DIFF
--- a/src/components/Nav.astro
+++ b/src/components/Nav.astro
@@ -148,9 +148,6 @@ li {
 nav.dark li {
     color: var(--color-white);
 }
-li + li {
-    margin-left: 0;
-}
 [aria-current="page"] {
     font-weight: 700;
 }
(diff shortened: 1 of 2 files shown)


delucis:
This looks good Fred! Agree that this is a clearer organisation of those links.

One note: the “Guides” and “Documentation” links will both end up in exactly the same place (`/en/getting-started/`). Maybe that’s fine as it gives people two different pathways to the docs, but just wanted to ensure that’s intentional.

tony-sull (approved):
LGTM 🚀 I'll think a bit more about the Guides and Documentation links

Eventually we will replace Guides with a Tutorial but until then maybe the Documentation link actually goes straight to the Core Concepts page?

delucis:
@sarah11918 Has a tutorial _all_ but written, so if you were wanting one for this, I think all you need to do is sneeze?

PR #144, opened 2022-06-01
Fix code issue in "Server-Side Rendering with Astro" blog post

kidonng:
Using `await` in a function makes it a `async` function.

Since it is example code, this may look a little nitpicking, but I hope you don't mind it :)


tony-sull (approved):
Not a nitpick at all, great catch!  It may just be a blog post code sample, but it should work when copy/pasted 👍

PR #161, opened 2022-06-19
add stefenphelps.com to showcase

stefenphelps:


src/data/showcase/elian-codes.json
```diff
@@ -1,6 +1,6 @@
 {
-	"title": "Elian Codes",
```
tony-sull, line 2: Revert this file and we should be all set 👍

tony-sull (changes requested):
Thanks for opening this PR!  Added one comment to revert an unrelated file and we'll get this merged in 🎉

src/data/showcase/elian-codes.json
stefenphelps: ack! Will do. Thanks!

tony-sull:
Sorry for the delay here!  The Netlify deploys went a bit haywire and I wanted to double check why they were hanging before the merge :+1:

PR #173, opened 2022-06-29
Remove staff from astro.build contributor script

FredKSchott:
I think the project has come far enough that staff is less relevant than it once was! Core is big and awesome enough now that we can highlight core as the highest level on our homepage.


delucis:
Sounds good! Should this PR also run the script to update the data? Or should we add it to the data collection CI?

tony-sull (approved):
LGTM!

I went ahead and removed any mention of the `staff` role and re-ran the script to update `contributors.json`

PR #179, opened 2022-07-04
Fix clicking the search button not focusing the searchbox

zadeviggers:
For some reason, the way it was implemented didn't work (in Firefox at least). This should fix the issue.


tony-sull:
Thanks for getting the ball rolling here, @zadeviggers!  Turns out this was a bit more involved to update the markup similar to what @delucis mentioned above

Closing this one out to make way for [#190](https://github.com/withastro/astro.build/pull/190) with the markup fix

PR #180, opened 2022-07-04
Replace `clsx` with built-in `class:list` directive

### LINKED ISSUE #126: Dogfooding: Use Astro's `class:list` instead of clsx
I noticed the website is using `clsx` instead of Astro's `class:list`, which should be feature equivalent. That'd be a good opportunity to dogfood our stuff

delucis:
Closes #126

As suggested by @Princesseuh, we can use `class:list` and remove the dependency on `clsx`: dogfooding _and_ one less dependency!


tony-sull (approved):
lgtm 🚀

PR #188, opened 2022-07-06
Migrating to @astrojs/image

tony-sull:
[x] Showcase images
[x] Themes catalog
[x] Sponsor avatars (was a custom `sharp` script :eyes:)


src/components/Picture.astro
```diff
@@ -0,0 +1,41 @@
+---
```
tony-sull, line 1: Temporary!  This will be moved directly into `@astrojs/image`

scripts/update-sponsors.mjs
```diff
@@ -22,39 +20,11 @@ function normalizeMember(member) {
 	}
 }
 
-async function cacheAvatar(member) {
```
tony-sull, line 25: avatars are now generated at build time

PR #189, opened 2022-07-08
Chore: SEO and Accessibility audit

tony-sull:
[x] fixes missing or unhelpful alt text
[x] uses passive event listeners where needed, with feature detection
[x] adds width & height props to the "Deploy to netlify" button


delucis (commented):
These changes LGTM, @tony-sull! I’m see the a11y issues reported by Lighthouse on the live site all addressed 🥳 

A few further key things flagged by Axe (https://deque.com/axe/devtools/) and Firefox’s built-in a11y panel:

#### Header

- The Astro logo and text aren’t `aria-hidden` although there is `sr-only` text in the link that wraps them.
image

- The chevrons next to the flyout menu items aren’t labelled. They probably should be as you need to hit enter as a keyboard user to reveal the flyout. Or, reveal the flyout on focus?
   image


#### Homepage

- the `<demo-illustration>` components both have `role="img"` but are missing `alt` text
- the SVG framework logos are missing `alt` text / should probably be `aria-hidden="true"` as they duplicate the framework names in the list
- the “devtools FM” link has no associated text/label because it’s only an SVG icon

#### Themes page

- `astro-icon` instances aren’t labelled. (The GitHub icon does have text associated with the link, so OK to use `aria-hidden` on that one)
   image image image

#### Integrations page

- Similar to themes: star & download icons aren’t labelled
- Integration logos also need `alt=""` I think

#### Blog index

- Twitter icons need `aria-hidden` — accessible text inside the link is available

tony-sull:
Oh geez, I hadn't realized Lighthouse missed so many accessibility warnings compared to Axe :eyes: Doing a second pass now for Axe warnings

PR #190, opened 2022-07-08
Updating search bar markup for proper accessibility

tony-sull:
Updates the search bar used on integrations to use a `<label>` for better accessibility and automatically focus the input on click


delucis (commented):
Looks good @tony-sull! Just spotted a couple of small things.

PR #215, opened 2022-07-28
Fix performance bug in `<not-marquee>` banner

### LINKED ISSUE #175: Optimize the banner animation
I haven't measured the performance of the banner using any tooling, but I have a relatively slow computer and I can clearly see that the banner is making the website way slower whenever it's on the screen and the animation is playing (the navbar dropdown hover animations and text selection become noticeably slower).
I'm not sure exactly why it's being so slow, but if you look at the actual CSS for the animation:
```css
@keyframes marquee {
  from { transform: translate3d(calc(100% * var(--i, 0)), 0, 0) }
  to { transform: translate3d(calc(100% * calc(var(--i, 0) - 1)), 0, 0) }
}
```
Maybe translate3d is expensive. I'm really just guessing at this point.

tony-sull:
Closes #175


orenelbaum:
It's still kind of slow on my pc, not sure why.

PR #216, opened 2022-07-28
Accessibility Improvements

Yan-Thomas:
I audited `astro.build` with the same test suite used in the docs repo:
- Screen reader navigation using NVDA (Duplicate/missing announcements, unclear or unnecessary landmarks, non-descriptive text, etc).
- Keyboard-only navigation.
- Mobile navigation with/without screen reader and keyboard-only.
- High Contrast/Forced Colors mode (Elements boundaries visibility, unclear active elements, styling conflicts, etc).
- HTML/ARIA semantics.
- Color contrast.

Per-commit, additional comments to clarify decisions and reasons (plus images!):

[Add custom search placeholder prop](https://github.com/withastro/astro.build/commit/57d54e467cd5a14338934da8e0d302f3364db957)
Placeholder text needs to be descriptive and not only use the same label as the input name/type as a way to add more clarity and avoid duplicate announcements. For this reason, I used "Search integrations/themes" as the placeholder text.

[FlyoutMenu accessibility improvements](https://github.com/withastro/astro.build/commit/7f024fc6fd332e7e56c9e792af280c28165f5eac)
- `<details>` elements don't need additional "open" labels since they are already announced as collapsable.
- I removed the navigation landmark inside the collapsible menu, it isn't a different navigation context but rather collapsed links. Having multiple, non-labeled navigation landmarks is bad UX for people who navigate using landmark shortcuts.
- Added an invisible outline (it becomes visible in high contrast) so the collapsed links' boundaries are clear.
- Removed the alt attribute in `<a>` tags.

[Improve cards accessibility](https://github.com/withastro/astro.build/commit/815778e203650ac8fb190e9e30f5d13a82239798)
- Wrapped the cards in lists, so screen reader users know how many themes/integrations/showcases are. 
- Added `role="listitem"` for each card, I wanted to use `<li>` but this caused some CSS issues I couldn't fix, but this role has good compatibility across screen readers so I think it's fine.
- Added an invisible outline for high contrast to know the card's boundaries.

| PR  |  astro.build  |
| ------------------- | ------------------- |
| image | image|

[Add descriptive anchor link text](https://github.com/withastro/astro.build/commit/99ad8944a82a4b18163285f0e9d9a08ab8499db4)
- NVDA was announcing headings as "Automatic Partial Hydration, number", this happens because of the "#" anchor link being part of the text. I changed the anchor a bit so it is announced as "Automatic Partial Hydration, section anchor link"

[Highlight selected pill in high contrast](https://github.com/withastro/astro.build/commit/86040ee2c3e7c95a2d4da0de05fb3b1d0262c243)
- In high contrast mode the currently selected pill was not being highlighted, I added a forced-colors media query that fixes the problem using a system color:

| PR  |  astro.build  |
| ------------------- | ------------------- |
| image | image|

[Fix PixelLink contrast in 404 page](https://github.com/withastro/astro.build/commit/985321ec41825b8f8776bdc8f692c2bc869500a5)
- How I will return home if I can't see "Return Home"? 🤣

| PR  |  astro.build  |
| ------------------- | ------------------- |
| image | image|

[Use descriptive text in the Twitter link](https://github.com/withastro/astro.build/commit/b3117aeebc429b06ec58dd8aefa22529a69415db)
- NVDA announced the Twitter link only was "@astrodotbuild", which isn't clear and could mean any social that uses "@" for profiles, so I changed it for "Twitter".

[Indicate Aside boundaries in high contrast](https://github.com/withastro/astro.build/commit/f2941ffa40e227eac08762f0f46429dae9f2031c)
- Same trick I used for the cards for the same reason.

[Rework landing heading structure and remove unnecessary labels](https://github.com/withastro/astro.build/commit/1fb1dde0be8119a8cd215df2d3d7aecd0eda309e)
- In the commit below I removed h1 from the header nav, this prompted me to restructure the heading levels to be more accurate to the page visual structure (h2 for each of Astro's important features). This should also improve SEO a bit.

[Remove Nav duplicate announcements, unnecessary label/heading, and add primary navigation label](https://github.com/withastro/astro.build/commit/8aba209edef52f190e7fe84458109b2d9754ddc3)
- Removed some unnecessary labels, like using `aria-label` and `aria-hidden` at the same time (aria-hidden just removes the element from the accessibility tree so other roles and labels are ignored).
- Changed `<h1>` for `<span>` since we don't need a heading for header links and this just worsens screen reader users' navigability.

[Indicate DemoIllustration boundaries in high contrast](https://github.com/withastro/astro.build/commit/35c21c9c8a365090d53f91fdbdb50f5ece625481)
- Same old trick, but here's a comparison:

| PR  |  astro.build  |
| ------------------- | ------------------- |
| image | image|

Note: The header is white instead of yellow in the PR because of the anchor link change, now the anchor is an adjacent element rather than part of the heading so it's properly colored in high contrast as a heading and not as a link.

[Add Footer secondary navigation landmark and remove copyright from the list context](https://github.com/withastro/astro.build/commit/7434a140c2ed3bf5664863274a19b6113b7c2125)
- The copyright text was being indicated as a list item, this incorrectly makes the user believe you have 5 links to navigate to, rather than 4.
- The footer should be considered a navigation landmark since it's globally available on the page and has different links from the header (that's why "Connect with Astro" isn't its own navigation landmark, but more like a call to action)

[Change blog tab order and add high contrast border](https://github.com/withastro/astro.build/pull/216/commits/30e579e2c6e762b42f207f14d80d7ec9c062321d)
- Changed the order in which the elements are focused (currently the Twitter link is being focused before the blog post itself)
- Removed the unnecessary "Read more from" text to improve navigability. Now screen reader users can easily skim through the blog posts without needing to wait for "Read more from" to be read or have to use skip text shortcuts.
- Added a border indicating the blog post boundaries for high contrast users (in this case I'm using a border rather than an outline because the focus outline was being mixed with the boundaries outline)

| PR  |  astro.build  |
| ------------------- | ------------------- |
| image | image|

### DIFF
--- a/src/components/Aside.astro
+++ b/src/components/Aside.astro
@@ -20,6 +20,8 @@ const { type = 'tip', title, class: className } = Astro.props;
 		background: var(--color-dawn);
 		color: var(--color-midnight);
         border-radius: var(--corner-md);
+		/* Outline only visible for high contrast users */
+        outline: 1px solid transparent;
 	}
 
 	aside > :global(* + *) {
--- a/src/components/FlyoutMenu.astro
+++ b/src/components/FlyoutMenu.astro
@@ -20,20 +20,22 @@ const hasDescriptions = items.some(({ description }) => !!description);
 <details class:list={['flyout', align, { light: invert }]}>
     <summary>
         {title}
-        <Sprite pack="heroicons-solid" name="chevron-down" size={20} aria-label={`Open ${title} menu`} />
+        <Sprite pack="heroicons-solid" name="chevron-down" size={20} aria-hidden="true" />
     </summary>
 
     <div>
-        <nav class:list={[{ descriptions: hasDescriptions }]} aria-label={title}>
+        <ul class:list={[{ descriptions: hasDescriptions }]}>
             {items.map(({ title, href, description }) => (
-                <a href={href} alt={title} class="subtle">
-                    <span>{title}</span>
-                    {description && (
-                        <p>{description}</p>
-                    )}
-                </a>
+                <li>
+                    <a href={href} class="subtle">
+                        <span>{title}</span>
+                        {description && (
+                            <p>{description}</p>
+                        )}
+                    </a>
+                </li>
             ))}
-        </nav>
+        </ul>
     </div>
 </details>
 
@@ -77,6 +79,8 @@ const hasDescriptions = items.some(({ description }) => !!description);
         opacity: 0;
         transition: opacity 150ms ease;
         will-change: opacity;
+        /* Outline only visible for high contrast users */
+        outline: 1px solid transparent;
     }
 
     details.left div {
@@ -87,7 +91,7 @@ const hasDescriptions = items.some(({ description }) => !!description);
         transform: translateX(-85%);
     }
 
-    nav {
+    details ul {
         display: flex;
         flex-direction: column;
         gap: 2rem;
@@ -105,7 +109,7 @@ const hasDescriptions = items.some(({ description }) => !!description);
         opacity: 1;
     }
 
-    nav.descriptions {
+    ul.descriptions {
         width: 100vw;
         text-align: start;
         white-space: normal;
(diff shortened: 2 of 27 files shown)


tony-sull (approved):
This is excellent, thanks @Yan-Thomas!

PR #222, opened 2022-08-04
Fix `list-style: none;` being overriden on contributors list

zadeviggers:
I noticed that the contributors list looked like this on both firefox and safari on macOS, so I had a quick peek in the dev tools and noticed that it was being overridden by some sort of `.prose` class. I think there may be some other issues with selectors, given that there's a `no-prose` class on the list itself, but this will fix the immediate issue.
Screen Shot 2022-08-05 at 12 08 40 AM


delucis (approved):
Thanks for catching that! 💜

PR #241, opened 2022-08-15
Update integrations.json

bronze:
An SEO plugin was missing from the  "performance+seo" category


tony-sull:
Looks like the nightly build that updates integration data failed a few times, this is fixed now 🎉

PR #248, opened 2022-08-19
🐛  fix: Fix broken link of `position-component` integration

trbnhck:
**What this PR does**
This PR fixes the bug, described down below, and replaces the dead link with an updated one.

-----

**Describe the bug**
When clicking on the `position-component` integration it redirects to a dead URL

**To Reproduce**

- Go to Page https://astro.build/integrations/css+ui/
- Click on `position-component`

**Screenshots**

https://user-images.githubusercontent.com/53407297/185571545-9a47889f-0c99-425f-a8d1-e59a7fbb70e6.mov

**Expected Behavior**

The URL should redirect to https://github.com/sacarvy/astro-components/tree/main/packages/position-component instead of https://github.com/sacarvy/astro-components/tree/main/position-component, since the integration was moved to a sudirectory.

```diff
- https://github.com/sacarvy/astro-components/tree/main/position-component
+ https://github.com/sacarvy/astro-components/tree/main/packages/position-component
```

### DIFF
--- a/src/data/integrations.json
+++ b/src/data/integrations.json
@@ -1043,7 +1043,7 @@
             "text": "View position-component on NPM"
         },
         "url": {
-            "href": "https://github.com/sacarvy/astro-components/tree/master/position-component",
+            "href": "https://github.com/sacarvy/astro-components/tree/main/packages/position-component",
             "text": "Learn more about position-component"
         },
         "downloads": 317,


delucis:
Thanks @trbnhck! We pull this data automatically from npm, so this would need to be updated in the integration itself (and republished to npm) for us to fix this: <https://github.com/sacarvy/astro-components/blob/69cd681bade236ffae0d46dff3368b66f5c756fb/packages/position-component/package.json#L25>

Perhaps @sacarvy would accept a PR to fix that.

sacarvy:
thanks a lot, I updated

PR #324, opened 2022-10-03
Tailwind right margin class affecting nav

### LINKED ISSUE #323: Navbar wonky between 640px and 767px width
Someone in Discord noticed over the weekend that the responsive nav bar has a small problem range.

Between 640px and 767px screen width, the menu items are displayed horizontally when opened/closed by the hamburger icon which overlaps some of the text.

Screenshot 2022-10-03 08 55 43

sarah11918:
Closes #323 

Issue: For a small range of screen widths, the hamburger menu opens the nav bar horizontally, but also overlaps with the first menu item. 
Screenshot 2022-10-03 08 55 43

Changing the Tailwind class from `-mr-2` to `mr-2` fixes the hamburger overlap, but when the item is focused, the outline still touches the first nav menu item: (Good enough, though, when no focus is visible.)

Screenshot 2022-10-03 09 14 05

This PR changes the Tailwind class to `mr-3` which shifts the entire hamburger and its outline far enough so as not to interfere with the menu items, but when closed, is maybe a little too large for what you'd ideally want?

Screenshot 2022-10-03 09 15 26
Screenshot 2022-10-03 09 15 41

Anyway, do with this what you will!

### DIFF
--- a/src/components/Header.astro
+++ b/src/components/Header.astro
@@ -64,7 +64,7 @@ const stars = formatter.format(monorepo.stars)
                         <Icon name="full-logo-dark" height={32} class="-mb-1" aria-hidden="true" />
                         <span class="sr-only">Astro</span>
                     </a>
-                    <div class="-mr-2 flex items-center md:hidden">
+                    <div class="mr-3 flex items-center md:hidden">
                         <button
                             type="button"
                             id="menubtn"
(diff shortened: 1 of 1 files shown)


tony-sull:
@sarah11918 oh great catch, I hadn't noticed the header layout in that ~650px width!  Tweaked it a bit so the menu button handles the right margin better in that size, just need to double check it in a preview deployment

PR #354, opened 2022-10-18
Fix failing AXE error

natemoo-re:
Only `li` elements can be children of `ul`. This removes an errant `footer` element.


delucis (approved):
LGTM!

Only used in one place currently and no visual changes caused by this fix:

| Before | After |
|---------|-------|
| image | image |

tony-sull (approved):
lgtm

PR #363, opened 2022-10-24
Add `@astrojs/og` to blocklist

natemoo-re:
`@astrojs/og` is super experimental and we probably don't want to publicize it yet

Also updates variable names to more inclusive `allowlist` and `blocklist`

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,8 +1,9 @@
 name: Nightly Build
 
 on:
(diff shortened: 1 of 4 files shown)


delucis (approved):
LGTM! 💜 

This doesn’t actually remove `@astrojs/og` just yet, so we might want to trigger the data collection action once this is merged (or just wait for the nightly run, may not be super urgent.)

matthewp (approved):

PR #365, opened 2022-10-25
New theme card style

itsMapleLeaf:
Update theme cards to match the neon border style on other parts of the site

Web capture_25-10-2022_11550_localhost


itsMapleLeaf:
I decided not to do the dark theme thing, 'cause the rest of the similar pages weren't doing it either

tony-sull (approved):
good call on sticking with the light theme for now, looks good :+1:

PR #371, opened 2022-10-27
New theme submit page

itsMapleLeaf:
image


src/pages/themes/submit.astro
```diff
@@ -0,0 +1,105 @@
...
+            placeholder="My Awesome Theme"
+            required
+        />
+        <div class="w-fit self-start">
+            <Field label="Free">
+                <Switch name="free" checked data-free-switch />
+            </Field>
+        </div>
+        <InputField
+            label="Live demo URL"
```
tony-sull, line 67: `type="url"` should give us basic browser data validation for the URLs

src/pages/themes/submit.astro
```diff
@@ -0,0 +1,105 @@
...
+            name="liveDemoUrl"
+            placeholder="https://example.com/theme-demo"
+        />
+        <InputField
+            label="Public repo URL (required for free themes)"
```
tony-sull, line 72: Same here for `type="url"`

src/pages/themes/submit.astro
```diff
@@ -0,0 +1,105 @@
...
+        <Button type="submit" theme="primary">Submit</Button>
+    </form>
+</Base>
+
+<script>
+    const freeSwitch =
+        document.querySelector<HTMLInputElement>('[data-free-switch]')
+    const repoUrl = document.querySelector<HTMLInputElement>('[data-repo-url]')
+
+    const updateRepoUrlRequired = () => {
```
tony-sull, line 95: Oh, nice!  Let's remove the "(required for free themes)" text too and toggle a red * here when it's required

src/pages/themes/submit.astro
```diff
@@ -0,0 +1,105 @@
...
+            </li>
+            <li>All NPM dependencies used in the theme are public</li>
+            <li>
+                Theme preview image or screenshot included (see below for
+                details)
+            </li>
+        </ul>
+    </section>
+
+    <form
```
tony-sull, line 37: Lets add their honeypot spam filtering too, it's not perfect but should block the dumb bots out there

https://docs.netlify.com/forms/spam-filters/#honeypot-field

src/pages/themes/submit.astro
```diff
@@ -0,0 +1,105 @@
...
+        method="post"
+        class="max-w-screen-sm flex flex-col items-center gap-4 mx-auto py-8 px-4"
+        data-netlify="true"
```
tony-sull, line 40: Lets add a custom success page that points to `/themes/submit/success`

https://docs.netlify.com/forms/setup/#success-messages

src/pages/themes/submit.astro
```diff
@@ -0,0 +1,105 @@
...
+<Base title="Submit a theme">
+    <PageTitle>
+        <Fragment slot="title">Submit a theme</Fragment>
+        Thank you for sharing your themes with the Astro community! Choose to make
+        your theme free or paid. All payments go directly to you, we don't handle
+        the sale or take a cut.
+    </PageTitle>
+
+    <section class="max-w-screen-md mx-auto px-4">
+        <h3 class="text-3xl mt-8 mb-4">Best practices</h3>
```
tony-sull, line 21: This should be an `<h2>` since there isn't another one on the page

tony-sull (changes requested):
It's looking great! Really like the image upload component, nice to see there was a clean no-JS fallback there

Added a few comments and extra features we should add before shipping

A few design ideas I was testing out below, not a great way to include them in a review but I can push them up if we want to keep them :+1:

Screen Shot 2022-10-28 at 2 32 23 PM

src/pages/themes/submit.astro
itsMapleLeaf: Didn't know that was a thing! :+1:

src/pages/themes/submit.astro
itsMapleLeaf: Ah yeah, forgot about that too

src/pages/themes/submit.astro
itsMapleLeaf: That's a really neat technique actually

itsMapleLeaf:
For the image input, I converted it to a Preact component, because the logic for it was getting tangled and hard to work with :sweat_smile: I also updated it so the "Please choose a file" bubble actually shows if you try to submit without one, which doesn't show when it has the `hidden` attribute

tony-sull (approved):
looks good!  Crazy that Netlify Functions doesn't support mutli-file uploads

PR #386, opened 2022-11-02
Update data sources

aleksa-codes:
Some of the links on the Integrations page were linking back to https://astro.build, changed them to lead to a reference in the doc (for prism and markdown-component) and GitHub readme (for agnosticui-astro).

Prism and legacy markdown-component are mentioned in multiple places, so I tried to pick the reference that gives the most useful information about them.

### DIFF
--- a/src/data/integrations.json
+++ b/src/data/integrations.json
@@ -17,7 +17,7 @@
             "text": "View @astrojs/prism on NPM"
         },
         "url": {
-            "href": "https://astro.build",
+            "href": "https://docs.astro.build/en/reference/api-reference/#prism-",
             "text": "Learn more about @astrojs/prism"
         },
         "downloads": 136099,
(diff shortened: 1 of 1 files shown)


delucis:
Thank you @aleksa-codes! Unfortunately this data is automatically scraped every night, so your manual changes will be overwritten next time that runs.

Instead we will need to update the URLs in `package.json` in https://github.com/withastro/astro

Would you be up for that?

aleksa-codes:
> Thank you @aleksa-codes! Unfortunately this data is automatically scraped every night, so your manual changes will be overwritten next time that runs.
> 
> Instead we will need to update the URLs in each integration’s `package.json` file in https://github.com/withastro/astro
> 
> Would you be up for that?

@delucis My bad, I thought it was only for the repo stars. I will gladly do that! Have you maybe checked the links I added? Just to clarify if chose good ones, thank you.

aleksa-codes:
I've made PR hopefully it gets approved, thank you! Though I haven't found where the link for `agnosticui-astro` is coming from.

delucis:
Thanks! `agnosticui-astro` is a third-party package, so would need updating there. I think [this is the relevant package.json](https://github.com/AgnosticUI/agnosticui/blob/master/integrations/agnosticui-astro/package.json) on GitHub, but don’t know if they’re accepting PRs.

PR #392, opened 2022-11-06
fix: use updated Wow URL in API content example

ejdrien:
If someone were to try this code snippet for themselves after November 28, 2022, they might be disappointed because it wouldn't work. Therefore it is necessary to update the URL to a new version made by the author which will continue to work after November 28, 2022.


delucis (approved):
Wow! Thanks for updating that.

PR #405, opened 2022-11-15
Theme details JSON support

tony-sull:
Adds support for full theme details content in our JSON pipeline

> Still a work in progress, may need to rename a few variables to match up with the Netlify Forms -> PR script


src/data/themes/index.ts
```diff
@@ -1,22 +1,35 @@
...
+
+async function resolveImage(src: string) {
+    if (!(src in allImages)) {
+        throw new Error(`[loadThemes] "${src}" image not found! Does it exist in /src/data/themes/images?`)
+    }
+
+    const mod = await allImages[src]()
+
+    return mod.default
+}
```
itsMapleLeaf, line 32: The PR will use image URLs that are uploaded to Netlify from the form. Will that cause an error here?

src/data/themes/index.ts
```diff
@@ -1,22 +1,35 @@
...
 import type { Link, Theme } from '../../types.js'
 
 interface ThemeData {
     title: string
     description: string
-    image: { src: string; alt: string }
+    fullDescription?: string
+    image: { src: string; alt: string; }
+    images?: { src: string; alt: string; }[]
     repoUrl: Link
```
itsMapleLeaf, line 11: This should be optional now

src/data/themes/index.ts
```diff
@@ -1,22 +1,35 @@
...
     npmUrl?: Link
     demoUrl?: Link
+    links?: Link[]
     categories: string[]
     featured?: number
     slug: string
     stars: number
```
itsMapleLeaf, line 18: This should also be optional (and always was to begin with 😅)

src/data/themes/index.ts
```diff
@@ -1,22 +1,35 @@
...
+    tags?: string[]
```
itsMapleLeaf, line 19:
```suggestion
    tags?: string[]
    PREVIEW?: true
```
It might also be a good idea to use a zod schema to parse the theme JSON so it's _definitely_ the shape we expect

src/data/themes/index.ts
tony-sull: Can the form submission webhook commit images as part of the PR?  I'm not sure if there's any guarantee how long the images are available on Netlify's CDN, especially if we delete the form submission once we merge in each PR

src/data/themes/index.ts
tony-sull: Oh yeah that's a little confusing, this is actually the type *after* data comes out of the `update-themes.mjs` script.  Its defaulted to an empty array there, but I'll go ahead and mark it as optional here just in case

PR #412, opened 2022-11-18
Update defined-networking.json

jasikpark:
gotta get that perf - https://hyperdrive-speedometer.netlify.app/#site-7826272 `www.defined.net` is the main homepage, `defined.net` redirects to that


delucis (approved):
Haha, I don’t think that should actually impact the perf score but might be wrong. Either way, LGTM!

jasikpark:
from what i've seen in speedcurve, it does

PR #415, opened 2022-11-20
Fixed the alphinejs link to point to correct page.

sanathusk:
Fixed the alphine js link to point to correct page.


delucis (approved):
Thanks!

PR #434, opened 2022-11-29
Add theme submission fields for tools and keywords

itsMapleLeaf:
https://user-images.githubusercontent.com/19603573/204640938-9678473a-3f93-477c-aedd-931d224d7e78.mp4

Without JS, these fallback to a multi-select and a regular input to put comma-separated tags. In either case, the form submission should have comma-separated tags in them


src/components/forms/TagInput.tsx
```diff
@@ -0,0 +1,174 @@
...
+                                tags.delete(tag)
+                                return new Set(tags)
+                            })
+                        }}
+                    />
+                </div>
+                <input
+                    class="flex-1 outline-none bg-transparent"
+                    placeholder="Add tags (separated by commas or spaces)"
+                    onKeyDown={(event) => {
```
tony-sull, line 150: This should probably also run for onblur as well, tabbing out of the input box without hitting enter first doesn't add the keyword

tony-sull (changes requested):
Nice!  Just one quick change for keyboard Tab support in the keywords input

tony-sull (approved):

PR #453, opened 2022-12-06
Adds `astro-webfinger` integration and extra rel="me" links

tony-sull:
Did someone say [IndieWeb](https://indieweb.org)?

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -3,12 +3,23 @@ import mdx from '@astrojs/mdx';
 import prefetch from '@astrojs/prefetch';
 import sitemap from '@astrojs/sitemap';
 import tailwind from '@astrojs/tailwind';
+import webfinger from 'astro-webfinger';
 import { defineConfig } from 'astro/config';
 
 // https://astro.build/config
 export default defineConfig({
   site: 'https://astro.build',
-  integrations: [tailwind(), image({ serviceEntryPoint: '@astrojs/image/sharp' }), prefetch(), mdx(), sitemap()],
+  integrations: [
+    tailwind(),
+    image({ serviceEntryPoint: '@astrojs/image/sharp' }),
+    prefetch(),
+    mdx(),
+    sitemap(),
+    webfinger({
+      instance: 'webtoo.ls',
+      username: 'astro',
+    }),
+  ],
   vite: {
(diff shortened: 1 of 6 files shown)


natemoo-re (approved):
Nice!

FredKSchott:
Screen Shot 2022-12-06 at 2 47 58 PM

possible that this error existed before, since I only see 2/4. But FYI @tony-sull !

FredKSchott:
I'm actually a bit lukewarm on this change, and probably would have objected if I'd gotten to it in time. But in the interest of experiments, can we add some link tracking to fathom see how often people end up clicking on these? 

After a few weeks, I'd then revisit to keep / remove based on the final click data.

bholmesdev:
For the record, also wasn't thrilled with adding Mastodon this early given the account (and community) is fresh. Agreed to Fred's analytics comment 👍

tony-sull:
> possible that this error existed before, since I only see 2/4. But FYI @tony-sull !

Off, yep I scrolled back to the launch of the site and it looks like the header never quite handled ~tablet-size right 

Fixing that now

natemoo-re:
I’d say Mastodon can go in the footer but not the header?

sarah11918:
Just my opinion but if the purpose of the links is to allow people to find /contact us where THEY are, then I'm not sure why we wouldn't include the link prominently with the others. I'm seeing a lot of people leave Twitter and move to Mastodon, so that's where they are. 

I'm not sure people really have expectations of the (any) "corporate/org" account other than that's where they can keep up with news/announcements or send a message to us? The performance art is simply a bonus when it exists. As long as we expect to keep people minimally informed and reply when necessary ... I don't know why we wouldn't advertise that we're there.

bholmesdev:
@sarah11918 So I understand that, but I don't think prominent badges are _just_ about meeting people where they are. It's also the way to showcase who WE are, the quality of our content, and (admittedly) to show off vanity metrics for engagement. 

We're currently crushing it on GitHub, Discord, and Twitter, but our Mastodon has far less quality content and engagement. So, it feels premature to showcase this to others curious about Astro. If we _really_ pump out quality and find a community on Mastodon, I'm all for it. But I don't think we're there yet

PR #454, opened 2022-12-06
Add /on links for socials

natemoo-re:
Adds `/on/[platform]` social links as client-side redirects.

Allows our Mastodon social links to all be verified and users to find us more easily!

### DIFF
--- a/src/pages/on/[platform].astro
+++ b/src/pages/on/[platform].astro
@@ -1,0 +1,21 @@
+---
+import type { Link } from '../../types.js'
+import site from '../../data/site.js'
+
+interface Props {
+    link: Link
+}
+
+export function getStaticPaths() {
+    return site.socialLinks.map((link) => ({
+        params: { platform: link.name },
+        props: { link }
+    }))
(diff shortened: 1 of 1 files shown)


src/pages/on/[platform].astro
```diff
@@ -0,0 +1,21 @@
...
+        params: { platform: link.name },
+        props: { link }
+    }))
+}
+
+const mastodon = site.socialLinks.find((link) => link.name === 'mastodon')
+const { link } = Astro.props
+---
+
+{mastodon && <link rel="me" href={mastodon.me ?? mastodon.href} />}
```
Comment (tony-sull) [inline src/pages/on/[platform].astro:20]: Will the `??` break smoke tests in Node 14?

tony-sull (approved):
looks good!  One note just to double check this won't hit smoke testing

src/pages/on/[platform].astro
Comment (natemoo-re) [reply src/pages/on/[platform].astro:20]: I don't think so because `.astro` is compiled with TypeScript!

PR #455, opened 2022-12-07
remove mastodon from main header nav

tony-sull:
Follow-up to #453, includes Mastodon in the site footer but not the main navigation header


FredKSchott:
Screen Shot 2022-12-12 at 9 05 24 PM

still seeing it in the header!

PR #459, opened 2022-12-07
New theme details page and paid theme support

tony-sull:
[x] adds a new theme details page for every theme
[x] adds a new `/themes/submit` form, no more GitHub issue templates 🚀 
[x] updates the list of supported theme collections


delucis:
Love this!

Nit and I’m not quite sure why this happened but looks like a styling change on the homepage here:
| Production | This branch |
|---|---|
| image | image |

Reason I noticed was I was wondering if this change merits something additional on the homepage. We currently call out integrations & themes as part of the “ecosystem”:

image

That said, that could be a later change once we’ve collected more themes to show off.

tony-sull:
> Nit and I’m not quite sure why this happened but looks like a styling change on the homepage here:

I can't actually tell where the original break here came from 🤔 The background was causing an overflow on mobile, just pushed a fix so the overflow-hidden doesn't stop the background grid from fading out on desktop

natemoo-re (approved):

FredKSchott:
Screen Shot 2022-12-07 at 2 07 07 PM


This "Astro v1.0" badge is out of date, want to update it with something short and sweet about the themes catalog? Bonus if you can track how many people click it!

FredKSchott:
Screen Shot 2022-12-07 at 2 09 28 PM

nit: this template looks a bit crowded on my screen, would be better to either go for 2 columns (instead of 3), or handle it more like how Framer does 3-col layouts which is more abstract: 


Screen Shot 2022-12-07 at 2 10 42 PM

FredKSchott (approved):
LGTM design-wise! didn't get a chance to look at code so will leave that to others

PR #466, opened 2022-12-08
Add showcase submit form

itsMapleLeaf:
Available at `/showcase/submit`


src/pages/showcase/submit.astro
```diff
@@ -0,0 +1,190 @@
...
+
+            <Field label="Preview images (max 8MB combined)">
+                <ImageInput
+                    name="images"
+                    label="Preview image"
+                    required
+                    client:idle
+                />
+
+                <div class="grid md:grid-cols-2 w-full gap-4 pt-3">
```
tony-sull, line 112: We only actually collect one image for showcase, we can remove the additional image inputs and update the field label to "Preview image" or similar

src/pages/showcase/submit.astro
```diff
@@ -0,0 +1,190 @@
...
+            />
+
+            <InputField
+                label="Website name"
+                name="websiteName"
+                placeholder="My Awesome Website"
+                required
+            />
+
+            <TextAreaField
```
tony-sull, line 147: (change from the current form) We never actually used the description, maybe we just remove it now?

src/pages/showcase/submit.astro
```diff
@@ -0,0 +1,190 @@
...
+
+            <InputField
+                label="Public website URL"
+                name="websiteUrl"
+                type="url"
+                placeholder="https://my.awesome.website"
+                required
+            />
+
+            <div class="w-full">
```
tony-sull, line 162: We don't use this in the showcase today, open to collecting it though if you have an idea for a future update!

src/pages/showcase/submit.astro
```diff
@@ -0,0 +1,190 @@
...
+                </label>
+                <TagSelect
+                    client:idle
+                    id="tools"
+                    name="tools"
+                    options={toolOptions}
+                />
+            </div>
+
+            <div class="w-full">
```
tony-sull, line 174: Same here, we don't use keywords in the showcase but curious how we may use it down the road.  I could see it being interesting metadata when we have proper search 🤔

src/pages/showcase/submit/error.astro
```diff
@@ -0,0 +1,31 @@
+---
+import Button from '../../../components/Button.astro'
+import PageTitle from '../../../components/PageTitle.astro'
+import Base from '../../../layouts/Base.astro'
+---
+
+<Base title="Theme submission failed">
```
tony-sull, line 7: nit: update the messaging for showcase vs. theme submissions

tony-sull (changes requested):
realize I didn't include much direction in the task related to what info we need to collect for showcase submissions, sorry about that!

Flagged a few mentions of themes here and marked a couple input fields as either not needed or interesting uses for later when we use SSR 👀

tony-sull (approved):
Looks great!  Good riddance, github issues...

PR #479, opened 2022-12-13
bump node version to 18

itsMapleLeaf:
Node 18 is the current LTS, and Node 14 is doing things I don't understand (related to theme/showcase submissions)


itsMapleLeaf:
Submission now works, and I clicked through the whole rest of the site on the preview and didn't notice any issues 👀

PR #495, opened 2023-01-05
UI: Searchbard styles improvement on Themes and Integrations

hijuliancode:
Hello, I would like to propose a change for the search engine on the integrations page.

The current browser is not visible and I think we could improve it by adding that it is always visible and also the possibility of making focus when pressing the / key, inspired by the documentation page.

- The searchbar is now full width
- The searchbar is focused when the "/" is pressed

Feedback would be appreciated and I can help with other UI improvements.

Screenshot 2023-01-05 at 00 32 16

### DIFF
--- a/src/components/Search.astro
+++ b/src/components/Search.astro
@@ -19,7 +19,7 @@ const inlineStyle = `
 ---
 
 <section>
-    <label>
+    <label for={id}>
         <input type="text" id={id} placeholder={placeholder} />
         <div>
             <Sprite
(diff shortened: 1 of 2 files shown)


tony-sull:
Sorry this PR sat open for so long, @hijuliancode!

We're actually wrapping up a revamp to these pages *including* a proper search and filter feature that took some refactoring to work well with SSR :eyes:

Hope you don't mind I'm going to close out this PR since the page will be overhauled soon, **but** the new designs will have an always-visible search bar.  I'll also open add a task to hook up the `/` key to focus search feature, that's very clever!

Thanks for raising this **and** opening a PR with suggested fixes, sorry again for the oversight as we've been heads down on the new pages!

PR #497, opened 2023-01-06
Fix invisible a11y issue in lighthouse

TheOtterlord:
Lighthouse showed an accessibility issue with the background to text contrast ratio on a specific element. The issue didn't really exist, because the `Starfield` component floats behind, applying a gradient with a dark enough color not to cause any issues. Lighthouse can't detect this, so we give a "fallback" background to the header which exists behind the `Starfield` component. 

TLDR: The end result is the same, but the single accessibility issue on the homepage is fixed

image

### DIFF
--- a/src/components/Starfield.astro
+++ b/src/components/Starfield.astro
@@ -9,7 +9,7 @@ const { subtle, noGradient } = Astro.props as Props;
 
 <div class:list={[
     "pointer-events-none",
-    "absolute inset-0 border-t-[3.5rem] border-slate-900",
+    "absolute inset-0 -z-50 border-t-[3.5rem] border-slate-900",
     subtle && "bg-[length:100%_175%]",
     noGradient ? 'bg-slate-900' : 'bg-gradient-to-b bg-no-repeat bg-purple-800 from-slate-900 to-purple-800'
 ]}>
--- a/src/components/Starfield.astro
+++ b/src/components/Starfield.astro
@@ -9,7 +9,7 @@ const { subtle, noGradient } = Astro.props as Props;
 
 <div class:list={[
     "pointer-events-none",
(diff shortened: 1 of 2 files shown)


TheOtterlord:
context: https://discord.com/channels/830184174198718474/846469231176056853/1060994646798499981

MoustaphaDev:
https://pagespeed.web.dev/report?url=https%3A%2F%2Fdeploy-preview-497--astro-www-2.netlify.app%2F
image

MoustaphaDev (approved):

src/components/Starfield.astro
```diff
@@ -9,7 +9,7 @@ const { subtle, noGradient } = Astro.props as Props;
 
 <div class:list={[
     "pointer-events-none",
-    "absolute inset-0 -z-50 border-t-[3.5rem] border-slate-900",
```
delucis, line 12: Looks like removing the negative `z-index` here means the starfield covers text in the page header on other pages, e.g. https://deploy-preview-497--astro-www-2.netlify.app/blog/

Maybe instead of removing here you could set a lower `z-index` in the homepage Hero component?

TheOtterlord: Ah yes, thanks for catching this xD
I tried setting the lower `z-index` for the Hero component, but it breaks the HeroCard interactions on the homepage.

I'll see what else I can try to get this working.

TheOtterlord: As I published that update, I found a fix by instead increasing the `z-index` on that same Hero component. As far as I could see, everything now plays nicely and stacks in the correct order

<details>
<summary>Screenshots</summary>
<br/>
Homepage
<br/>

<br/>
Other page headers (e.g. blog)
<br/>

<br/>
Lighthouse score
<br/>

</details>

delucis (approved):
Looks great — thanks @TheOtterlord! 🙌

PR #498, opened 2023-01-09
Fix 404 monitoring

delucis:
I noticed that all 404 events in Fathom are being reported as for the `/404/` path rather than a useful path like we get for docs. My guess is that this because we don’t set a canonical URL for 404 pages in docs but this is set for astro.build.

This PR sets the canonical URL to `null` on the 404 page. The SEO component we’re using still emits the canonical tag but with the value an empty string, so we need to test if Fathom will ignore that as I would hope.

### DIFF
--- a/src/components/BaseHead.astro
+++ b/src/components/BaseHead.astro
@@ -8,7 +8,7 @@ export interface Props {
     title?: string
     description?: string
     image?: string | ImageMetadata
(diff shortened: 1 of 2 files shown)


tony-sull (approved):
Oof, guess defaulting canonical URL has a downside after all!  lgtm

PR #501, opened 2023-01-13
Content collections refactor

bholmesdev:
- Bump dependencies to beta for Astro core, `@astrojs/image`, `@astrojs/mdx`, and `@astrojs/tailwind`
- Migrate blog post parser to a collection schema
- Refactor `getAuthor(id)` to better use schemas
- Update `BaseHead` and `BlogPost` templates to use content collections
- Replace explicit `zod` dependency with imports from `astro/zod`


src/components/Mention.astro
```diff
@@ -1,12 +1,10 @@
...
+import { getAuthor } from '../data/authors'
+import Avatar from './Avatar.astro'
+const { name } = Astro.props
 
-const mention = await getAuthor(name);
-if (!mention) {
-    throw new Error(`Could not find mentioned user "${name}"!`)
-}
+const mention = getAuthor(name)
+if (mention instanceof Error) throw Error
```
itsMapleLeaf, line 7: looks like this should be `throw mention`

Although, this looks the same as `getAuthor` throwing the error itself, but with extra steps :thinking: `getAuthor` doesn't even appear to return `Error`

bholmesdev: Ah yep sorry, used to return the error object, and it now throws internally. Good to remove this

tony-sull:
@bholmesdev Netlify is pulling the Node version from `.nvmrc`.  For the 2.0 beta are smoke tests using Node 16 now?

src/pages/blog/_utils.ts
```diff
@@ -1,48 +1,8 @@
-import path from 'node:path';
```
tony-sull, line 1: bye bye, boilerplate code 🚀

bholmesdev:
TODO: bump to non-beta versions and merge!

tony-sull (approved):
lgtm 💰

PR #526, opened 2023-02-02
Remove redundant entries for @astrojs/cloudflare adapter

AirBorne04:
Currently on the integrations page for adapters there are 4 instances of cloudflare adapters (astro-cloudflare, astro-cloudflare-node, astro-js-cloudflare), all linking to the main repo and they should probably removed in favour of the `@astrojs/cloudflare` package.

### DIFF
--- a/scripts/npm.mjs
+++ b/scripts/npm.mjs
@@ -73,8 +73,14 @@ export async function searchByKeyword(keyword, ranking = 'quality') {
         total = results.total
     } while (total > objects.length)
 
-    return objects.reduce((acc, next) => {
-        acc.set(next.package.name, next)
-        return acc
-    }, new Map())
+    return objects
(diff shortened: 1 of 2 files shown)


tony-sull:
Oh that's a weird one, thanks for filing this PR!  Looks like an NPM account re-published multiple copies of our cloudflare integration? 🤔 

The `src/data/integration.json` file is actually rebuilt nightly, I'll push an update here that modifies the `/scripts` used by the nightly job 👍

AirBorne04:
Ah ok, i was wondering if it generated or manual 😀

tony-sull:
Yep, it's generated nightly but in desperate need of better docs!  Think I actually have a partially done README stashed, today seems like a good day to dust that off - this repo was originally private and we missed adding a good readme when making it public

PR #539, opened 2023-02-17
Delete taep96.moe from /showcase

taep96:
I rebuilt my website on another framework, so removing it from the /showcase page.


delucis:
Thanks @taep96! I think I also see you in Astro's Discord server. If you ever feel like sharing your reasons for moving framework in the `#feedback-ideas` channel or want to DM me, please do. It can be really helpful to hear from folks how their Astro experience was.

Happy coding! 🙌

delucis (approved):
Thanks! Build failed but I reran on Netlify and it built fine so just a fluke it looks like.

PR #545, opened 2023-02-22
a11y fixes

delucis:
- Add explicit true value to `aria-hidden` where missing

- Fix color contrast of links on [theme pages](https://deploy-preview-545--astro-www-2.netlify.app/themes/details/portfolio/)
    | Before | After |
    |---|---|
    | image | image |

Also noticing that the Quill editor we’re using for the theme submission form is missing accessible labelling for its toolbar. There’s been [an open PR to fix that on their end](https://github.com/quilljs/quill/pull/3639) for nearly 6 months but seems maybe they’re not interested. Might be good to re-evaluate choice of text editor for the new site to see if there’s a more accessible alternative?

### DIFF
--- a/src/components/forms/ImageInput.tsx
+++ b/src/components/forms/ImageInput.tsx
@@ -35,7 +35,7 @@ export default function ImageInput(props: {
                             {props.required && (
                                 <span
                                     class="text-red-500 text-base align-text-top"
-                                    aria-hidden
+                                    aria-hidden="true"
                                 >
                                     *
                                 </span>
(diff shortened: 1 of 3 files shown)


tony-sull (approved):
looks good to me!  Yeah I'm not super happy with that open bug on Quill for their toolbar either, unfortunately at the time this was the *most* accessible MD->HTML editor I came across

@itsMapleLeaf might have a better solution for the new form, definitely something we can revisit 👍

PR #551, opened 2023-03-05
chore: use https links

imba-tjd:


tony-sull:
Thanks @imba-tjd!  We're literally just about to hit deploy on a new site redesign, I manually ported over these fixes to #552 to avoid any kind of crazy merge conflicts since the entire codebase is being updated 🚀

PR #558, opened 2023-03-06
Update Houston images

natemoo-re:


jasikpark:
Can `@astrojs/image` be used to generate Webp images w/ the pngs?

jasikpark:
image
oh, i guess they could all be webp?

natemoo-re:
Ehh I don't know enough about our image setup. I leave that to @tony-sull! But I think we are just hard-coding `png` for consistency right now.

tony-sull:
Yep, we're pre-rendering `webp` images for now **but** will be switching to the new core image APIs as soon as possible!

doodlemarks:
thank you @natemoo-re!

PR #559, opened 2023-03-06
un-chop the lit-html logo

e111077:
~unfortunately I can't seem to run the site locally so hopefully it looks good.~ Noticed that the icons on the site are 32x32 so I made this 26x32. Hopefully taht looks ok.

Preview:

| before | after|
| - | - |
| image | image |


natemoo-re (approved):
Whoops, not sure what happened there. Thanks for the fix! ♥️

jasikpark (approved):

PR #560, opened 2023-03-07
Use larger community cloud image

delucis:
Update the image accompanying the community card on the new homepage to be 1.5x instead of 1x. I also slightly tweaked the position of a couple avatars that weren’t quite centred on the circle — designers please yell at me if I shouldn’t have 😅 

This also fixes the mobile layout, which didn’t quite match the Figma designs yet.

#### Before (desktop)
Before screenshot
#### After (desktop)
After screenshot

#### Mobile comparison
| Before | After |
|---|---|
| image | image |


tony-sull (approved):
🚀

PR #562, opened 2023-03-07
Home page optimisations

delucis:
- Smaller image sizes for the heavier website screenshots in the “Content-focused” section
- Smaller icon sizes for Vercel & AlpineJS (Vercel icon was 42 KB!!)
- Avoids some subtle CLS issues by adding `height`/`width` to a couple homepage images


src/pages/_components/ExampleShowcase.tsx
```diff
@@ -4,9 +4,9 @@ import Collapse from "~/components/Collapse.jsx"
 import { positiveMod } from "~/helpers/math.js"
 import agenciesImage from "../_assets/example-agencies.png"
```
tony-sull, line 5: You beat me to it!  Lets go ahead and convert these to webp as well, Lighthouse always prefers modern formats anyway

src/pages/_components/ExampleShowcase.tsx
```diff
@@ -168,6 +168,8 @@ export default function ExampleShowcase() {
 							</p>
 							<img
 								src={item.image.src}
+								width={item.image.width}
```
tony-sull, line 171: Lets mark these as with `loading="lazy"` and `decoding="async"` as well since these are always below the fold on an initial page load

I could be sold on leaving the first image with eager loading since it's higher priority 🤷

src/pages/_components/GetStarted.astro
```diff
@@ -20,6 +20,8 @@ import houston from "../_assets/houston.png"
 			data-src={houston.src}
 			data-src-hover={houstonHappy.src}
 			src={houston.src}
+			width={houston.width}
```
tony-sull, line 23: I've got a hunch webp would help with these images as well!  Same for lazy loading and async decoding, no reason to block on this image down the page

tony-sull (changes requested):
Very nice, thanks for jumping on this!  Added a few quick comments on a bit more optimization that I think might help while we're in this code

src/pages/_components/ExampleShowcase.tsx
delucis: I actually did try, but found these images had minimal gains (2–3 kB) for a visibly lower quality. So I stuck with compressing the couple that were seriously heavy.

src/pages/_components/ExampleShowcase.tsx
delucis: Done in 58fb0a0.

I think we’re cool to whack `lazy` on all of these given it’s lower down the page for now.

src/pages/_components/GetStarted.astro
delucis: Good hunch! These are 90% smaller with WebP 🎉

src/pages/_components/ExampleShowcase.tsx
tony-sull: Oh that's interesting!  Gotta love how all over the place different compression formats are!

tony-sull (approved):
Very nice, thanks Chris!

PR #568, opened 2023-03-08
fix: showcase page mobile view overflow

nattwasm:
## Description 🗒️

Showcase page has responsive issues with mobile view because of the background image overflowing on the x-axis. I fixed it with this merge request by adding hiding the overflow on the x-axis.

fix: #569

## Screenshots 📷

CleanShot 2023-03-07 at 17 05 16@2x

Before

CleanShot 2023-03-07 at 17 05 56@2x

After


tony-sull (approved):

tony-sull:
Looks like that may have been a regression this afternoon, thanks for catching this!

PR #570, opened 2023-03-08
Proxying custom fonts through a /fonts redirect with caching

tony-sull:
[x] adds a redirect in Netlify from `/fonts/*` to fonts-cdn.astro.build
[x] adds cache headers for `/_astro/*`, these are all hashed bundles that can safely have long cache expirations

## Download comparisons

I ran around a dozen load comparisons with and without the redirect + caching.  Looking at the `Obviously Regular` font there's a **big** drop in Content Download time since the fonts are cached on the same domain

**Before**
Screenshot 2023-03-08 at 7 35 49 AM

**After**
Screenshot 2023-03-08 at 7 36 18 AM

### DIFF
--- a/public/_headers
+++ b/public/_headers
@@ -1,3 +1,6 @@
 /assets/*
   cache-control: max-age=604800
   cache-control: public
+/_astro/*
+  cache-control: max-age=31536000
+  cache-control: immutable
--- a/public/_headers
+++ b/public/_headers
@@ -1,5 +1,5 @@
 /assets/*
-  cache-control: max-age=604800
+  cache-control: max-age=86400
   cache-control: public
 /_astro/*
   cache-control: max-age=31536000
--- a/public/_headers
+++ b/public/_headers
@@ -1,3 +1,8 @@
 /assets/*
-  cache-control: max-age=86400
+  cache-control: max-age=604800
   cache-control: public
+  stale-while-revalidate=86400
+/_astro/*
+  cache-control: max-age=604800
+  cache-control: public
+  stale-while-revalidate=86400
--- a/public/_headers
+++ b/public/_headers
@@ -1,8 +1,6 @@
 /assets/*
   cache-control: max-age=604800
   cache-control: public
-  stale-while-revalidate=86400
 /_astro/*
   cache-control: max-age=604800
   cache-control: public
-  stale-while-revalidate=86400
--- a/public/_headers
+++ b/public/_headers
@@ -1,6 +1,3 @@
 /assets/*
   cache-control: max-age=604800
   cache-control: public
-/_astro/*
-  cache-control: max-age=604800
-  cache-control: public
--- a/src/components/BaseHead.astro
+++ b/src/components/BaseHead.astro
@@ -75,3 +75,7 @@ const resolvedImage = {
 />
 
 <link rel="preload" href={Inter} as="font" type="font/woff2" crossorigin />
+
+<!-- image preloads -->
+
+<link rel="preload" as="image" href="/assets/noise.png" />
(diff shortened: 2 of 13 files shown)


src/components/BaseHead.astro
```diff
@@ -1,7 +1,9 @@
 ---
 import Inter from "@fontsource/inter/files/inter-latin-variable-wghtOnly-normal.woff2"
+import "@fontsource/inter/variable.css"
```
tony-sull, line 3: refactor: moving the CSS loading into the same file where preload headers are added

src/components/BaseHead.astro
```diff
@@ -58,18 +60,22 @@ const resolvedImage = {
...
 	as="font"
 	type="font/woff2"
 	crossorigin
 />
 
 <link rel="preload" href={Inter} as="font" type="font/woff2" crossorigin />
+
+<!-- image preloads -->
+
+<link rel="preload" as="image" href="/assets/noise.png" />
```
tony-sull, line 81: unrelated to fonts, noticed the noise image file could use a preload for a slight perf improvement

public/_headers
```diff
@@ -1,3 +1,6 @@
 /assets/*
-  cache-control: max-age=86400
+  cache-control: max-age=604800
+  cache-control: public
+/_astro/*
+  cache-control: max-age=604800
   cache-control: public
```
delucis, line 6: Are we able and do we want to set headers for `/fonts/*` if we’re redirecting? We do have `public, max-age=604800, stale-while-revalidate=86400` set in the fonts-cdn so may not be needed. It would also be interesting to circle back to the CDN to see if there improvements we’d want to make.

delucis (commented):
Thanks @tony-sull! I’m curious why this proxy changes things performance?

public/_headers
```diff
@@ -1,3 +1,6 @@
 /assets/*
-  cache-control: max-age=86400
+  cache-control: max-age=604800
+  cache-control: public
+/_astro/*
+  cache-control: max-age=604800
```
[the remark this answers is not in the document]
tony-sull: Yep, that'll work here too!  If I remember right CDNs will treat `immutable` the same as `public` but it will tell browsers to not even bother checking if the file has changed as long as the cache hasn't expired yet

public/_headers
tony-sull: Yeah let me try not adding cache headers here at all, I didn't see cache headers in the `font-cdn` code base but it was late when I was playing around with this!

tony-sull:
> I’m curious why this proxy changes things performance?

The main benefit is to avoid an extra DNS lookup and HTTPS connection.  Proxying it lets Netlify cache the font files and treat them the same as any other file in `/public`, when the browser circles back to download the fonts it's already connected to the server and doesn't have to go back through the whole TLS handshake process

I was actually a bit surprised how much time this picked up.  On a not-the-best internet connection I was seeing the "Waiting for Server Response" time the same but the actual "Content Download" step dropped from ~250ms for one font file to ~20ms.  It was just enough that the font `preload` tag finished before rendering started and I never see the hero font flash at all!

public/_headers
tony-sull: Yep, that's what I missed!  I was only looking for headers in the actual edge function response

We're getting good cache hits with the redirect + the cache headers from fonts-cdn so we should be all set here :+1:

delucis (approved):
I’ll call out that I’m not able to reproduce the differences in loading time for the fonts using this technique vs direct use of CDN URLs, but it’s also not _worse_, so I’m fine with this.

The caching change for `/_astro/` is good as are the other small refactors 👍

PR #572, opened 2023-03-08
Use Netlify CLI to enable redirects in dev

delucis:
This PR changes the `pnpm start` and `pnpm dev` scripts to run Astro’s dev server via `netlify dev`. By doing this — and moving redirect and header config to `netlify.toml` — we get an equivalent dev environment to what runs on Netlify.

Motivation for this change was that after switching to a `/fonts/*` proxy redirect to our CDN, fonts are currently 404-ing in development.


tony-sull (approved):
You read my mind, looks good to me!

PR #580, opened 2023-03-08
Notification banner a11y fixes

delucis:
- Adds an accessible label reading “Hide Banner” to the X icon button in the notification banner

- Tweaks the colour contrast which was technically failing WCAG AA contrast requirements. New version on the left, old version on the right:

    image

    (The gradient change is equivalent to changing lightness in the HLS colour from 63% to 60%)


Yan-Thomas (approved):
Nice catch @delucis, LGTM! 💜

tony-sull (approved):
👨‍🍳 💋

PR #581, opened 2023-03-08
Replacing the resources menu with a <details> element

tony-sull:
Accessibility improvements here, replacing the custom Resources menu in the desktop layout with a `<details>` element

Added a bit of javascript that automatically hides menu once it loses focus when you either Tab or click outside of the menu


Yan-Thomas (approved):
Can't confirm if the Safari arrows issue was fixed as I don't have anything from Apple, but a11y LGTM, tested with NVDA. 💜

delucis (commented):
Trying to narrow down what’s doing this, but on a narrow viewport I’m seeing:

1. Click hamburger to open menu
2. Click resources/chevron to open resources submenu
3. Click hamburger to close menu <- nothing happens!
4. Click hamburger to close menue again <- menu closes

🤔

tony-sull:
@delucis Nope that's definitely not a you thing!  I didn't get a crash but I'm seeing the menu animate itself closed/open immediately on that first no-op click, there must be a bug in the Solid component that handles the menu 🤔 

Taking a look now, too bad that bug hit here.  Replacing the JS menu with a fancy no-JS (or progressively enhanced) solution was next on my list to investigate 😅

delucis (approved):
LGTM! Diagnosed some odd menu behaviour but seems independent from this PR — thanks for tackling this Tony!

tony-sull:
double confirmed, that menu bug is already in main - will start a separate PR to track down that fix ASAP!

PR #583, opened 2023-03-08
:bug: Fixes a click handler loop in the mobile menu

tony-sull:
In the mobile menu, if you click on the Resources dropdown then try to close the menu both elements end up handling the event and call `setVisible(!visible())` twice, flipping it to closed and then back open immediately

This updates the focus-lost handler to ignore the even if the mobile menu button is the new focus target


delucis (approved):
Can confirm this is now working as expected in FF and Chrome. Thanks for find a fix so quickly @tony-sull!

PR #586, opened 2023-03-08
Fix Netlify redirects

delucis:
Try changing order in netlify.toml in case it’s significant


delucis:
Nope — order is not relevant.

PR #592, opened 2023-03-09
Add skip to content link

Yan-Thomas:
Adds a "Skip to content" link!
 
Note: 
- Couldn't find a good way to vertically align with the navbar, it would depend on some calculations that would not work depending on if the notification bar is visible or not.
- Base white button used custom gradient/background styles that were incompatible with transform/drop-shadow for some reason a Tailwind noob can't understand, therefore I "forked" them. Also, removed the transition to avoid a super animated skip link, removing `transition: none` brings it back if we want.

### DIFF
--- a/src/components/HeaderNav.astro
+++ b/src/components/HeaderNav.astro
@@ -1,15 +1,12 @@
 ---
 import AstroLogo from "./AstroLogo.astro"
 import ResourcesMenu from "./ResourcesMenu.astro"
-import SkipLink from "./SkipLink.astro"
 import SocialLinks from "./SocialLinks.astro"
 ---
 
 <a href="/" title="Home" data-home-link>
 	<AstroLogo />
 </a>
-
-<SkipLink />
 
 <script>
 	import { getElement } from "~/helpers/dom.js"
--- a/src/components/HeaderNav.astro
+++ b/src/components/HeaderNav.astro
@@ -1,12 +1,15 @@
 ---
 import AstroLogo from "./AstroLogo.astro"
 import ResourcesMenu from "./ResourcesMenu.astro"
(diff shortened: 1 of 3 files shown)


delucis:
Quick question: what was the motivation behind placing this after the Astro logo homepage link? I’d have expected this first in tab order.

Yan-Thomas:
> Quick question: what was the motivation behind placing this after the Astro logo homepage link? I’d have expected this first in tab order.

@tony-sull suggested this, based on Netlify's site navigation. I'm not sure if there's more to it. Since the Astro homepage and notification bar (which can be dismissed) are I'd say, important (and not many) links, I think it doesn't hurt the user experience.

delucis:
The goal is to make it easy to bypass blocks of content repeated across pages, so I’d say that definitely include the home logo link and _maybe_ the banner too even though it’s dismissible?

The [WCAG note on this](https://www.w3.org/TR/2012/NOTE-WCAG20-TECHS-20120103/G1) specifically suggests this being the _first_ element on the page:

> The objective of this technique is to provide a mechanism to bypass blocks of material that are repeated on multiple Web pages by skipping directly to the main content of the Web page. The first interactive item in the Web page is a link to the beginning of the main content.

Yan-Thomas:
> The goal is to make it easy to bypass blocks of content repeated across pages, so I’d say that definitely include the home logo link and maybe the banner too even though it’s dismissible?

Yeah, I agree, especially if there isn't a bigger motive for this.

tony-sull:
I don't have a strong opinion either way, happy for it to be the first focusable element if that's the WCAG recommendation!

delucis (approved):
This looks good to me! Thanks for adding this @Yan-Thomas — amazing work as always 💜

PR #594, opened 2023-03-09
Update the Netlify logos

lorenzolewis:
Updates the Netlify logos per their new rebrand: https://www.netlify.com/press/#logos https://www.netlify.com/blog/netlify-new-logo/

Similar PR in docs repo: https://github.com/withastro/docs/pull/2817

I don't like how the new logo has a lot more vertical height than the old one. This is more apparent in the sponsor position when it's scaled down even more.


tony-sull:
Looks like the build failure is related to a security rule slipping environment variables, we'll get that fixed on our end!

lorenzolewis:
No prob, give me a shout if anything needs rebased after

PR #596, opened 2023-03-09
improve collapses on main page

itsMapleLeaf:


src/components/Collapse.tsx
```diff
@@ -1,37 +1,40 @@
...
+		})
+		ro.observe(inner!)
+		onCleanup(() => {
+			ro.disconnect()
+		})
 	})
 
 	onMount(() => {
-		inner.style.position = "absolute"
+		inner!.style.position = "absolute"
```
tony-sull, line 24: nit: safer to wrap this in an `if(inner)` just in case something crazy happens and the node wasn't found (same below)

itsMapleLeaf: I actually don't think so... if the node isn't found, it would break silently, and on an error, at least we're aware that a mistake is made, and it doesn't take the entire site down, just this callback fails to run 😅

tony-sull: Ah ha, I was just guessing it'd be unhandled and break JS.  Just tried it locally and the Collapse throws an error but doesn't actually break anything else :tada:

tony-sull (approved):
buttery smooth 👍

PR #597, opened 2023-03-09
WIP: Adds the "Submit your site" banner in the showcase

tony-sull:
This adds the "Submit your site" banner to the Showcase


Yan-Thomas (approved):
LGTM, great work @tony-sull! 

Though, I initially had difficult to find where the "Submit your site" banner was, perhaps we should consider moving this to the top or end of the page, so its location will be more obvious? Or using a more contrasting background color? - Could totally just be a "me" thing, so it's non-blocking comment.

delucis (approved):
Agree with Yan’s comment — not sure if this was just because I _expected_ this to be at the end of the page like on the old site but took me a second. That might be OK though and it looks very pretty!

Other than that, looks good once we have the discussion set up!

tony-sull:
Discussion is actually live now, I just forgot to remove that note in the PR!

That's great feedback on the banner being lost in the mix.  That was the original design so I'm going to merge it as-is to push submissions to the GH discussion, but will start a thread tomorrow to see if we want to revisit the position 👍

PR #599, opened 2023-03-10
add prefetch links

itsMapleLeaf:
Added rel="prefetch" wherever I thought it made sense

### DIFF
--- a/src/components/Footer.astro
+++ b/src/components/Footer.astro
@@ -113,8 +113,8 @@ const groups: LinkGroup[] = [
 	<div
 		class="grid grid-cols-[1fr,auto] items-center gap-x-4 gap-y-16 py-8 opacity-60 md:grid-cols-[auto,1fr,auto] md:py-12"
 	>
-		<a class="link" href="/privacy" rel="prefetch">Privacy Policy</a>
-		<a class="link mr-auto" href="/terms" rel="prefetch">Terms of Service</a>
+		<a class="link" href="/privacy">Privacy Policy</a>
+		<a class="link mr-auto" href="/terms">Terms of Service</a>
 		<p class="col-span-2 text-center md:col-span-1">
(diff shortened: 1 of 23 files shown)


src/components/Pagination.astro
```diff
@@ -49,6 +49,7 @@ const collapsedPages = collapseRange(page, pages)
 					) : (
 						<a
 							href={link.href}
+							rel="prefetch"
```
tony-sull, line 52: let's skip this one for now, scrolling to the bottom of the page ends up triggering quite a few prefetches for prev/next + individual pages + terms/privacy

I do like the next/prev arrows though!  Those are probably a lot more likely to be clicked than a page number 

Screenshot 2023-03-10 at 12 41 29 PM

src/components/Footer.astro
```diff
@@ -113,8 +113,8 @@ const groups: LinkGroup[] = [
 	<div
 		class="grid grid-cols-[1fr,auto] items-center gap-x-4 gap-y-16 py-8 opacity-60 md:grid-cols-[auto,1fr,auto] md:py-12"
 	>
-		<a class="link" href="/privacy">Privacy Policy</a>
-		<a class="link mr-auto" href="/terms">Terms of Service</a>
+		<a class="link" href="/privacy" rel="prefetch">Privacy Policy</a>
+		<a class="link mr-auto" href="/terms" rel="prefetch">Terms of Service</a>
```
tony-sull, line 117: Lets skip these two for now too, they're probably so rarely clicked it's not worth the extra network requests when a user scrolls to the end

tony-sull (changes requested):
Nice!  I had tested this once before launch and was seeing a crazy FOUC, not seeing it here now so not sure what was happening 🤷 

Just marked a couple that I'd vote we skip prefetch for now to help out anyone on a slower network

tony-sull (approved):
lgtm 💰

PR #601, opened 2023-03-10
Fix a11y in showcase submit component

delucis:
- Adds `alt=""` to mark Houston as decorative

- Wraps the submit component in `<li>` as it’s inside `<ul>`.

  Not sure if there‘s a better way given I think the intent was for this `<aside>` to be outside of the list flow. Maybe if the `<ul>` had `display: contents`, the `<aside>` could follow `<ul>` and the `<aside>` and all the `<li>`s follow the grid of a wrapper component? But I seem to remember `display: contents` messing up list item ARIA role maybe? So went with this simple solution — NWTWWHB!


Yan-Thomas (approved):
Perfect @delucis, LGTM!

> Not sure if there‘s a better way (...)

Yeah, I did some research and couldn't find a better solution, and `display: contents` sometimes cause weird bugs, so NWTWWHB!

PR #603, opened 2023-03-11
SPA

natemoo-re:
Just here for the preview link

### DIFF
--- a/src/layouts/MainLayout.astro
+++ b/src/layouts/MainLayout.astro
@@ -26,7 +26,7 @@ const { ...head } = Astro.props
 			<NotificationBanner uid="astro-210">
 				<a href="/blog/astro-210/" rel="prefetch">Astro version 2.1 just shipped!</a>
 			</NotificationBanner>
-			<div id="nav" class="sticky top-0 z-20 max-h-screen">
+			<div id="nav" class="sticky top-0 z-20 max-h-screen" data-persist>
 				<Header>
 					<HeaderNav />
 					<HeaderDrawerContent slot="drawer" />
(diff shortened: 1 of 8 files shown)


src/layouts/MainLayout.astro
```diff
@@ -55,5 +55,27 @@ const { ...head } = Astro.props
...
+						if (cache.has(href)) continue
+
+						promises.push(
+							fetch(new URL(href, window.location.toString())).then((res) => {
+								cache.add(href)
+							}),
+						)
+					}
+					return Promise.race([
+						new Promise((resolve) => setTimeout(resolve, 100)),
```
jasikpark, line 74: what is all of this??? the animation is so nice tho

natemoo-re: This is a hacky way to preload the incoming CSS, if there is any

natemoo-re: Removed! Not needed.

src/layouts/MainLayout.astro
```diff
@@ -62,25 +62,7 @@ const { ...head } = Astro.props
...
-							}),
-						)
-					}
-					return Promise.race([
-						new Promise((resolve) => setTimeout(resolve, 200)),
-						Promise.all(promises),
-					])
-				},
-			})
+			listen()
```
jasikpark, line 65: hey, listen

tony-sull:
Closing this out for now but I'm *very* interested to reopen this PR for a few more preview builds when I have a chance to really put it through its paces!

I want to really hammer on the performance a bit since this would add a SPA-mode with DOM diffing, very curious to see how it stacks up!

PR #604, opened 2023-03-12
Fix 404 on blog post

lorenzolewis:
Is this a sign I have too much time on my hands? 😅


lorenzolewis:
Ah no problem then. It was on a local build that I was working with, didn't think to check the netlify config. Can confirm it redirects on the live site.

tony-sull:
Going to close this one out and let the original post fly, really appreciate you taking the time to file a PR for it just in case!

We had to actually revert an update last week that moved our scripts to use Netlify CLI for `netlify dev` which would have picked up those redirects locally.  Time to revisit fixing the issue in that PR so we can improve the local dev experience!

lorenzolewis:
I didn't even realise you could setup the netlify redirects to work locally as well. Do you have a link to that PR I could look at taking some inspiration from for the Tauri website?

delucis:
Here you go @lorenzolewis! https://github.com/withastro/astro.build/pull/572

lorenzolewis:
Thank you! Do you have an issue open or something with a bit context of why it was reverted?

lorenzolewis:
Coolio! I'll pay around with it on our beta site and see if anything wonky happens. Thanks again!

PR #605, opened 2023-03-13
Fix: accessible-astro-components URL change

Lakelimbo:
Was just randomly browsing the integrations and found that they changed the URL.


jasikpark (approved):
lgtm

tony-sull:
That build break is related to an issue with how our deploys are setup in the new codebase!  Looking into a fix with #606 that unblocks the build break when private env variables aren't defined in community PRs

tony-sull:
That's more like it, community PRs unblocked!

tony-sull (approved):
thanks for catching this!  Still working on a nightly script that updates data from NPM and github 😅

Lakelimbo:
> thanks for catching this! Still working on a nightly script that updates data from NPM and github 😅

no problem! And thank you for the fix as well! 😄

PR #606, opened 2023-03-13
fix: lazy validate env variables in the theme submit route

tony-sull:
This moves the check for `DISCORD_` environment variables to a runtime check when the form is submitted

## Why?

Netlify PR preview builds from community members won't include private environment variables, breaking the build immediately

### DIFF
--- a/src/process-env.ts
+++ b/src/process-env.ts
@@ -5,13 +5,15 @@ const schema = z.object({
 	SUBMISSION_DISCORD_WEBHOOK_THREAD_ID: z.string(),
 })
 
-const result = schema.safeParse(import.meta.env)
-if (!result.success) {
-	throw new Error(
-		`Missing environment variables: ${result.error.issues
-			.map((issue) => issue.path.join("."))
-			.join(", ")}`,
-	)
-}
+export function env() {
+	const result = schema.safeParse(import.meta.env)
+	if (!result.success) {
+		throw new Error(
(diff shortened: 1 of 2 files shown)


src/process-env.ts
```diff
@@ -5,13 +5,15 @@ const schema = z.object({
 	SUBMISSION_DISCORD_WEBHOOK_THREAD_ID: z.string(),
 })
 
-const result = schema.safeParse(import.meta.env)
-if (!result.success) {
```
delucis, line 9: Instead of removing this wholesale maybe we move it behind a `import.meta.env.CONTEXT === 'production'` check? This would then only error for production builds (could maybe warn in other situations?)

Docs: https://docs.netlify.com/configure-builds/environment-variables/#read-only-variables

itsMapleLeaf (approved):
Good change, I'm personally cool with this

src/process-env.ts
tony-sull: Oh yeah that's interesting!  I try to avoid depending on Netlify's built-in variables to avoid issues when they don't exist in `astro dev`, but that actually makes sense in this case

tony-sull: Turns out the fallback code gets a little wonky here checking `import.meta.env.CONTEXT` in addition to using `safeParse` to check for errors

Going to leave the easier fix here for now to unblock community PRs and open add it to the list of backlog tasks to revisit

PR #610, opened 2023-03-15
Bringing back a GitHub Action to update themes & integrations

tony-sull:
Adding a new GitHub Action cron job to update our themes and integrations catalog

The action will run every Monday morning, opening a PR with updates for:

- GitHub stars and author details for published themes
- NPM data for existing integrations
- adds newly published NPM packages that use the `withastro` and `astro-component` keywords
- removes packages that no longer exist on NPM (usually because they were deprecated)


delucis:
In case it‘s helpful, we use https://github.com/peter-evans/create-pull-request in docs for nightly PRs.

src/pages/themes/submit/index.astro
```diff
@@ -64,7 +63,7 @@ const errorResponse = (formData: FormData, errors: z.typeToFlattenedError<typeof
 }
 
 if (Astro.request.method === "POST") {
-	// convert astro request to a node-fetch request
+	// convert astro request to a fetch request
 	// hopefully we won't have to do this in the future 🥲
 	const formData = await new Request(Astro.request.url, {
 		method: Astro.request.method,
```
itsMapleLeaf, line 69: This makes me nervous! 😂 Does theme submission still work as expected?

scripts/github.mjs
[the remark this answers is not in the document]
tony-sull: 💯 agree, this would be a lot easier to maintain in TS!  Will leave that for a follow-up PR after confirming this *actually* runs successfully in prod (debugging/testing actions is a pain)

src/pages/themes/submit/index.astro
tony-sull: Yikes, yeah I didn't catch that when I was cleaning up the `node-fetch` uses!  I'm going to roll that back, the GitHub Action config is set to use Node v18 so there isn't really a need to switch the main project off `node-fetch`

tony-sull: Reverted back to `node-fetch`, the Github Action config will handle installing node 18 and using built-in `fetch()` separate from the main project

itsMapleLeaf (approved):
Nice use of a composite action!

tony-sull:
> Nice use of a composite action!

Borrowed that one from Docs, they have a stellar CI setup!  😍

PR #612, opened 2023-03-16
Add /wallpapers

natemoo-re:
Adds a `/wallpapers` page to download hi-res wallpaper assets 😍


jasikpark:
image

might want flex-wrap for the buttons? got weird wrapping in the buttons. also maybe a skeleton for the wallpaper would be good since it takes time to load so many xD

tony-sull (approved):
Looks good to me!

PR #613, opened 2023-03-16
WIP: migrating to the new `astro:assets` magic

tony-sull:
> ⚠️  This is a big migration and an experimental feature


tony-sull:
Showcase, themes, and the blog are all migrated to `astro:assets` 🎉 

Build times are a bit too slow for now without caching (~11 minutes in Netlify build runs) - going to close this out to revisit in the future with a caching solution

PR #614, opened 2023-03-16
switch to prettier for tailwind class sorting

itsMapleLeaf:
An upcoming PR will include some @apply in .css files, switching to prettier will allow sorting those as well

Also fixes lint script order so that prettier runs last, which removes the semicolons added by the import organizer


tony-sull (approved):
much `prettier`

PR #621, opened 2023-03-20
Migrating our prose styling to Tailwind's components layer

tony-sull:
This fixes (well, dodges) a potential specificity bug where the main Tailwind bundle is added to the page *after* a component's style

In this case, the reset for `p { margin: 0 }` was taking precedent over the `Prose.astro` component's `article > * + * { margin: 1.5rem }` style


itsMapleLeaf (approved):
💯 I kept thinking about doing this myself but never got around to it

PR #623, opened 2023-03-21
Adjusting prose content to max 70ch wide

tony-sull:
Updating the prose styling used in blog posts, theme detail pages, and generic pages like T&Cs and Privacy Policy

### DIFF
--- a/src/pages/blog/[slug].astro
+++ b/src/pages/blog/[slug].astro
@@ -51,7 +51,7 @@ const placeholderGradient = randomFromArray(gradients)
 >
 	{
 		coverImage ? (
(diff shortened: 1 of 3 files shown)


Yan-Thomas (approved):
LGTM @tony-sull!

I just have this small nit: the Twitter card isn't filling the 70ch width (or being centered). Considering it's an iframe it might not be seamless to change, and not something I'd say is blocking, so I'm fine with leaving this way.

image

tony-sull:
@Yan-Thomas Those Twitter embeds are the worst 😅  Their iframe actually breaks responsive layouts in weird ways by reading the rendered width once when it loads and inlining a `width: 500px` that can overflow when the window is resized...

Consider it a high-priority TODO to replace those any Tweet in our blog with a static representation of the tweet, we don't need all that JS in a blog post!

PR #640, opened 2023-03-31
Fixed InferGetStaticParamsType to actually use the Params type

marceloverdijk:
Fixed InferGetStaticParamsType to actually use the Params type in example code


delucis (approved):
Thanks for the thoroughness on this @marceloverdijk! This is a helpful change 🙌

PR #644, opened 2023-04-03
Use backticks to preserve CLI flag double-hyphen

markjaquith:
The double dash `--watch` was getting converted to an em-dash. ~~Putting a zero-width space between them breaks that up and~~ Wrapping in backticks makes the switch appear ~~(at least visually)~~ as two dashes.

Before:

CleanShot 2023-04-03 at 15 57 12@2x

After:

CleanShot 2023-04-03 at 17 03 53@2x

### DIFF
--- a/src/content/blog/astro-210.mdx
+++ b/src/content/blog/astro-210.mdx
@@ -18,7 +18,7 @@ We just released Astro 2.1, featuring:
 
 - **[Built-in Image Support](#built-in-image-support)**: (experimental) Automatic image optimization is landing in Astro core.
 - **[Markdoc Integration](#markdoc-integration)**: (experimental) A new content format from Stripe, optimized for large projects.
-- **[astro check -​-watch](#astro-check-watch)**: Use this flag to type check your `.astro` files during development.
+- **[`astro check --watch`](#astro-check-watch)**: Use this flag to type check your `.astro` files during development.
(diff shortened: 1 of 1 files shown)


src/content/blog/astro-210.mdx
```diff
@@ -18,7 +18,7 @@ We just released Astro 2.1, featuring:
 
 - **[Built-in Image Support](#built-in-image-support)**: (experimental) Automatic image optimization is landing in Astro core.
 - **[Markdoc Integration](#markdoc-integration)**: (experimental) A new content format from Stripe, optimized for large projects.
-- **[astro check --watch](#astro-check-watch)**: Use this flag to type check your `.astro` files during development.
+- **[astro check -​-watch](#astro-check-watch)**: Use this flag to type check your `.astro` files during development.
```
delucis, line 21: Thanks for pointing this out!

I’m not sure this is the ideal switch, because if anyone happens to try and copy paste this, the flag wouldn’t work and it wouldn’t be at all obvious why. (Admittedly copy/paste is not super likely in this context, but you never know!) I’d suggest wrapping this in backticks as it is technically some code:

```suggestion
- **[`astro check --watch`](#astro-check-watch)**: Use this flag to type check your `.astro` files during development.
```

markjaquith: Agree that having it be copy-pastable is better! (it's not copy-pastable in current form either, but it's visually obvious to anyone who knows about CLI flags).

delucis (approved):
Perfect — thanks again!

PR #666, opened 2023-05-05
output: 'server' vs 'ssr'

millette:
Did you mean 'server' instead of 'ssr' for the output config?


delucis (approved):
Oops, yes! 😅 

Thanks for catching that @millette 🙌

PR #673, opened 2023-05-16
remove simpler-astro-seo beacuse github repo doesn't exists anymore

JLarky:
https://github.com/perkinsjr/simple-astro-seo is 404

npm package has 9 weekly downloads


tony-sull (approved):
Thanks @JLarky!

Could have sworn we the weekly update script would catch these broken links. I'll have to double check how this one got missed

PR #678, opened 2023-05-23
New blog styles

itsMapleLeaf:
I left a test blog post for an example of how to use the various components I added + allows an explicit					 `publishDate: null` to have a blog post not show up in the index


src/content/config.ts
```diff
@@ -91,8 +91,9 @@ export const collections = {
 			description: z.string(),
 			publishDate: z
 				.string()
+				.transform((val) => new Date(val))
 				.or(z.date())
-				.transform((val) => new Date(val)),
+				.or(z.null()),
```
tony-sull, line 96: Probably worth leaving this to a separate PR if we need a draft mode for posts, with the page still being built by `src/pages/blog/[slug].astro` it'll end up in the production sitemap even though it's hidden on the blog index page

src/content/config.ts
```diff
@@ -91,8 +91,9 @@ export const collections = {
 			description: z.string(),
 			publishDate: z
 				.string()
```
tony-sull, line 93: This must have either been before `z.date({ coerce: true })` was a thing *or* before I realized the coerce option existed 😅 

We could probably simplify this to avoid the `string || date` + `transform` steps

tony-sull (changes requested):
Oh these new post components look great!

- not sure what the mobile layout design should be for the `Blockquote` (screenshot attached)
- would you mind saving a draft posts support for a separate PR?  There might be a few corner cases we need to figure out how it should work first

src/content/config.ts
itsMapleLeaf: I actually just updated it to exclude it from the sitemap haha

src/components/Blockquote.astro
```diff
@@ -0,0 +1,13 @@
+---
+import { Icon } from "astro-icon"
+---
+
+<figure class="relative border-y-[3px] border-astro-gray-500 py-4">
+	<Icon name="quote" class="absolute right-[calc(100%+1rem)] top-6 s-10" />
```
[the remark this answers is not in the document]
itsMapleLeaf: Hm... I'll poke Ximena on this

tony-sull (approved):
lgtm 💰

PR #680, opened 2023-05-29
Fix update-showcase script

delucis:
Escape the `after` query parameter in the graphql query that fetches discussion comments in the showcase script. We just hit a case where the value contains `=` and that is causing errors. This wraps the value in quotes to avoid that.

Triggered a run from this branch here: https://github.com/withastro/astro.build/actions/runs/5112550826/jobs/9190722693

Seems to be working as expected.


delucis:
This fix ran successfully so I'll merge — just a tiny change anyway.

PR #689, opened 2023-06-01
Migrating static images to `astro:assets`

tony-sull:
⚠️ don't let the file change count scare you off!  Almost all of that was refactoring where image files live, and pulling back in as many original source images as possible to replace the pre-built `.web` variants

Migrating most of our image optimizations to use `astro:assets`

[x] basic `<Picture>` component used for `.avif` and responsive size support
[x] Showcase card images
[x] Blog posts
[x] SEO share images (updated for imports only, no image optimizations)

### Note

This **does not** migrate over our themes catalog!  With the catalog supporting search and filtering on the server, the page is SSR'd and any card images would get built on-demand by sharp...not ideal.  There are a few ways we may be able to handle this, but they're more complicated and will be revisited in a follow-up PR


src/content/blog/experimental-server-side-rendering/index.mdx
```diff
@@ -8,6 +8,8 @@ authors:
   - matthew
 ---
 
+import netlifyDeployVideo from "./netlify-deploy-2.mp4"
```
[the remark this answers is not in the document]
tony-sull: @jasikpark Sorry, I totally missed this PR comment after letting the PR go stale during launch week!

Yep that's totally possible, at that time I had just missed the leading `_` so that the file is hidden from the content collections API

Importing a `.mp4` file ends up returning the hashed file path, basically `*.mp4?url` without needing the query param since it wouldn't make much sense for the import to return the raw binary content or anything like that

src/helpers/uniq.ts
```diff
@@ -0,0 +1,5 @@
+export function uniq(...values: string[]) {
+	const set = new Set()
+	values.forEach((value) => set.add(value))
+	return Array.from(set.values())
+}
```
itsMapleLeaf, line 5:
```suggestion
export function uniq<T>(...values: T[]): T[] {
	return [...new Set(values)]
}
```

src/components/Picture.astro
```diff
@@ -0,0 +1,45 @@
...
+	class?: string
+	src: ImageMetadata
+	alt: string
+	widths: number[]
+	sizes?: string
+}
+
+const { src, alt, widths, sizes = "100vw", class: className } = Astro.props
+
+const formats = uniq("avif", "webp", src.format) as ImageOutputFormat[]
```
itsMapleLeaf, line 17: After updating `uniq`, this cast can be removed

```suggestion
const formats = uniq("avif", "webp", src.format)
```

src/pages/themes/submit/index.astro
```diff
@@ -47,22 +49,6 @@ const schema = zfd
...
-
-	const params = new URLSearchParams({
-		errors: btoa(JSON.stringify(errors)),
-		data: btoa(stringifiedData.toString()),
-	})
-
-	return Astro.redirect(`/themes/submit?${params}`)
-}
-
 if (Astro.request.method === "POST") {
```
itsMapleLeaf, line 52: why was `errorResponse` removed?

src/types/env.d.ts
```diff
@@ -1 +1 @@
-/// <reference types="@astrojs/image/client" />
+/// <reference types="astro/client-image" />
```
itsMapleLeaf, line 1: I think we can delete this file, since there's already one at `src/env.d.ts`

itsMapleLeaf (changes requested):
Really nice to see this! It'll help with my case study work as well 😊 Just a question and an improvement to the `uniq` helper

tony-sull:
Well this one got to stale and the diff was getting out of hand, fresh `astro:assets` PR incoming!

PR #691, opened 2023-06-02
Highlighting recently added themes and integrations

tony-sull:
This adds a new "Recently Added" category for integrations which includes all integrations added in the last 28 days.  Three of the recent integrations are highlighted on the main `/integrations` page along with a link to view all recently added integrations


itsMapleLeaf (approved):

itsMapleLeaf:
It looks like you need to add `| undefined` to `title` on the card group component, and/or make it optional

PR #708, opened 2023-06-08
Removing moden-personal Theme

manuelernestog:
The Theme Astro Moder Personal Template Theme is now called Astrofy, and it was already submitted using the Astro.build website. I'm removing to avoid duplications on the theme list.


tony-sull:
Thanks @manuelernestog!

I just double checked our submissions and see the Astrofy theme waiting in our queue, it will go out with the next batch of submissions (should be early next week 🚀)

tony-sull (approved):

manuelernestog:
Thanks @tony-sull and thanks for your work. Really loving Astro and the path it's taking. 
Kudos for all the team 👏👏👏

PR #714, opened 2023-06-14
Using 16:9 aspect ratio for theme images

tony-sull:
In themes with only one image, the detail page should always show the image at 16:9 aspect ratio.

**Before** https://astro.build/themes/details/art-portfolio/ 
**After** https://deploy-preview-714--astro-www-2.netlify.app/themes/details/art-portfolio/


delucis (approved):
Code looks good to me! Do we have a specific page where this is happening we can look at?

delucis (approved):
Extra approved now I’ve seen the before/after 👍

PR #715, opened 2023-06-15
(WIP) Draft new release posts

TheOtterlord:
A work in progress / proof of concept script to generate a draft for release posts from the changesets in `withastro/astro`. It's got a lot of cleanup to do, if this is the direction we want for generating them.

Some notes:

- Needs retries (I'm avoiding any node modules to prevent waiting around for an install to finish, so I'll need to create this manually)
- Currently relies a lot on people using the first line of a changeset as a title/summary. What I might do is detect long first lines, and just put a placeholder title (maybe the integration name, as an example)
- Workflow in progress

Open to feedback


TheOtterlord:
<details>
  <summary>
    The anticipated Astro 2.6.4 release (I broke a couple of things to generate this as a test)
  </summary>
<pre><code>
---
title: "Astro 2.6"
description: ""
publishDate: ""
authors:
  - your-name-here
coverImage: ""
socialImage: ""
lang: "en"
---

{/* Introduce Astro 2.6 */}

If you already have Astro installed, you can upgrade it to 2.6 by running the `upgrade` command in your project (using your package manager of choice):

```
npm install astro@latest
pnpm upgrade astro@latest
yarn upgrade astro@latest
```

While you're at it, upgrade any `@astrojs/*` integrations and adapters you have installed, too!



## Let the `tailwindcss` PostCSS plugin load its config file itself. This changes the Tailwind config loading behaviour where it is loaded from `process.cwd()` instead of the project `root`.

If your Tailwind config file is not located in the current working directory, you will need to configure the integration's `configFile` option to load from a specific path:

```js
// astro.config.mjs
import { defineConfig } from 'astro/config';
import tailwind from '@astrojs/tailwind';
import { fileURLToPath } from 'url';

export default defineConfig({
	integrations: [
		tailwind({
			configFile: fileURLToPath(new URL('./tailwind.config.cjs', import.meta.url)),
		}),
	],
});
```

This change also requires a Tailwind config file to exist in your project as a fallback config is no longer provided. It is set up automatically during `astro add tailwind`, but if it does not exist, you can manually create a `tailwind.config.cjs` file in your project root:

```js
// tailwind.config.cjs
/** @type {import('tailwindcss').Config} */
module.exports = {
	content: ['./src/**/*.{astro,html,js,jsx,md,mdx,svelte,ts,tsx,vue}'],
	theme: {
		extend: {},
	},
	plugins: [],
}
```


## Rename options `config.path` to `configFile`, and `config.applyBaseStyles` to `applyBaseStyles`. If you are using these options, you need to migrate to the new names.

```diff
// astro.config.mjs
import { defineConfig } from 'astro/config';
import tailwind from '@astrojs/tailwind';

export default defineConfig({
  integrations: [
    tailwind({
-      config: {
-        path: '...',
-        applyBaseStyles: true,
-      },
+      configFile: '...',
+      applyBaseStyles: true,
    }),
  ],
});
```


## The deno adapter now supports npm package that require built-in node modules.

</code></pre>
</details>

matthewp:
Ah yeah, this is a great start, thank you! I think having this will push people (and reviewers) to have more succinct changesets and better titles. So I'm not worried about the long titles yet. It's a starting point that can be edited.

I can see a few enhancements in the future, like possibly populating `authors` from the authors of the changeset, but not important at the moment.

.github/workflows/blog.yml
```diff
@@ -0,0 +1,26 @@
+name: Draft minor release blog post
+description: Drafts a new minor release blog post
+
+on: workflow_dispatch
```
matthewp, line 4: This means it runs when you trigger it manually from the Actions tab, right?

TheOtterlord: yep!

TheOtterlord:
This is probably ready for review. There's obviously things I can still add to it (looking into adding authors now). I'm not sure what the best way to test the workflow is. I could create a test repository just to confirm it works.

PR #730, opened 2023-07-05
Fix misleading score transformation

endigma:
The old version weights the scores such that the performance of non-Astro frameworks is understated. Change rebases this scale to 0 which does generate a bit of "dead space" but more accurately shows the relative difference between the scores.


FredKSchott:
LGTM! Thanks @endigma for opening a PR to fix this!

PR #744, opened 2023-07-24
Update Astro blog avatar

doodlemarks:
updates blog avatar with latest version:

image

keep getting a collections error or id show the after but since it's just an image update 🤷


itsMapleLeaf (approved):
heck yeah byte reduction

bholmesdev (commented):
Hope you had a good avatime adding this!

PR #761, opened 2023-08-08
update case study gradient to be configurable

FredKSchott:


itsMapleLeaf (commented):
While this is a flexible way to do it, I think to avoid the "random color combinations all over the place" issue, we should make the option an enum of choices from our DS. WDYT?

[For reference, here's how the blog page randomly selects an option](https://github.com/withastro/astro.build/blob/05cf3639ee7da71c4641cfb216d6d46863de7053/src/pages/blog/%5Bslug%5D.astro#L36-L70)

FredKSchott:
I think we want this to always match the case-study brand colors, and not our own design system. These were Contenda's, and firebase would be their signature Blue + Yellow. Not sure matching to our brand colors makes sense? @doodlemarks can confirm.

itsMapleLeaf (approved):
I did not realize the first case study specifically used contenda's colors, it looked the same as a gradient in our DS 😅

In that case, this gets an approve from me 👍

PR #767, opened 2023-08-14
Restores astro-compress, astro-critters and astro-rome

NikolaRHristov:
This PR restores the missing packages as they are not considered new - https://github.com/withastro/astro.build/blob/26714efe56246196e43e0c2d557792fb4b557228/scripts/integrations.mjs#L18 and adds some icons.


NikolaRHristov:
image
image
image

Look good. 😁

PR #768, opened 2023-08-14
Showcase tweaks

delucis:
Follow up to #764

- Update the IDX.dev screenshot
- Marks a couple new sites as highlighted
- Rejigs some ordering with the new sites coming in

| Before | After |
|--------|--------|
| image | image |


tony-sull (approved):
Very nice!  Updated order looks great 🚀

PR #772, opened 2023-08-17
chore: Updating to latest core dependencies

tony-sull:
Updating the main build dependencies to latest releases

[x] `@astrojs/*` packages
[x] Tailwind & sharp (to match latest `@astrojs/*`)
[x] Inter and MD IO fonts

NOTE: `@astrojs/netlify` was left pegged to an earlier version to avoid https://github.com/withastro/astro/issues/7540


package.json
```diff
@@ -22,36 +22,35 @@
...
     "clsx": "^1.2.1",
     "gray-matter": "^4.0.3",
     "node-fetch": "^3.3.0",
     "quill": "^1.3.7",
-    "sharp": "^0.31.3",
+    "sharp": "^0.32.5",
     "slugify": "^1.6.5",
     "solid-js": "^1.6.10",
-    "tailwindcss": "^3.2.4",
+    "tailwindcss": "^3.0.24",
```
itsMapleLeaf, line 43: This got downgraded?

package.json
```diff
@@ -78,5 +77,5 @@
     "tiny-glob": "^0.2.9",
     "typescript": "^4.9.5"
   },
-  "packageManager": "pnpm@7.29.0"
+  "packageManager": "pnpm@8.6.0"
```
itsMapleLeaf, line 80: The latest version is `8.6.12`, although I'd consider future-proofing this to just `8.6` or even `8`. pnpm has a good major stability track record from my experience

src/components/BaseHead.astro
```diff
@@ -52,7 +52,7 @@ const resolvedImage = {
 
 <link
 	rel="preload"
-	href="/fonts/MD IO/Web/MDIO0.5-Regular.woff2"
+	href="/fonts/MDIO/Web/MDIO0.6-Regular.woff2"
```
itsMapleLeaf, line 55: Thank you for this, I kept getting annoyed that this font was wrong and I couldn't figure out why 🫠

src/pages/case-studies/[slug].astro
```diff
@@ -22,16 +16,15 @@ export async function getStaticPaths() {
 	}))
 }
 
-type Props = {
+export type Props = {
 	post: CollectionEntry<"caseStudies">
 }
 
-const { post } = Astro.props
+const { post } = Astro.props as { post: CollectionEntry<"caseStudies"> }
```
Comment (itsMapleLeaf) [inline src/pages/case-studies/[slug].astro:23]: Is this cast needed?

src/pages/blog/[slug].astro
```diff
@@ -17,11 +17,11 @@ export async function getStaticPaths() {
 	}))
 }
 
-type Props = {
+export type Props = {
 	post: CollectionEntry<"blog">
 }
 
-const { post } = Astro.props
+const { post } = Astro.props as { post: CollectionEntry<"blog"> }
```
Comment (itsMapleLeaf) [inline src/pages/blog/[slug].astro:24]: also this cast

package.json
tony-sull: Yep that's a great call, this was a byproduct of me fighting with Tailwind config not upgrading properly 🙃 

I'll set it to `8`, agreed pnpm is good about semver releases

package.json
tony-sull: Yeah it may not be needed, after the upgrade I ran into ridiculous build errors yelling about missing deps 3 layers deep in `autoprefixer` 😅 

This was my "just work!" fix, but let me revisit that.  It may have just been a bad lock file or local cache that was screwing with my build

src/pages/case-studies/[slug].astro
Comment (itsMapleLeaf) [reply src/pages/case-studies/[slug].astro:23]: No big either way if it's still broken 👍

Comment (tony-sull) [reply src/pages/case-studies/[slug].astro]: Ah ha, found it!  Looks like the language server is only automatically resolving `Astro.props` types if the `Prop` definition is at the top, not sure if that's expected or just a quirk of how it resolves the types + hoisted `getStaticPaths` definition

package.json
```diff
@@ -77,5 +77,5 @@
     "tiny-glob": "^0.2.9",
     "typescript": "^4.9.5"
   },
-  "packageManager": "pnpm@8.6.0"
+  "packageManager": "pnpm@8.0.0"
```
itsMapleLeaf, line 80: aa Sorry to block on this, but I'm pretty sure this'll use `8.0.0` exactly 😅 

If you haven't tried it, try `pnpm@^8` or `pnpm@^8.0.0`

If _that_ still doesn't work, then we'll use exactly `8.6.12` and save the trouble 🫠

tony-sull: `8.6.12` it is, the CI build blew up with `pnpm@^8.0.0` and the specs require a full `major.minor.patch` version 🙃

itsMapleLeaf (approved):

PR #774, opened 2023-08-18
migrating to `hybrid` rendering

tony-sull:
Moving over to `output: "hybrid"` since all but two of our routes are static

I actually caught a couple routes that were being SSR'd and didn't really need it, so chalk this up to a massive performance overhaul that will blow Lighthouse out of the water.


itsMapleLeaf (approved):
beautiful 💖

PR #779, opened 2023-08-21
Migrating to `astro:assets`

tony-sull:
> Follow-up to the now stale #613

This migrates most of the site to use `astro:assets` 🎉  We had been using static, pre-optimized images for everything which has gotten *very* long in the tooth for our repo size

Sections of the site migrated with this PR:
[x] blog posts & author avatars
[x] case studies
[x] Showcase catalog
[x] Themes catalog*

* Themes are now using `src/assets` **but** images are still pre-optimized for now.  A future PR will revisit options for supporting themes which use images both during SSG (theme detail pages) and SSR (search & filter)


tony-sull:
Going to shelf this for a bit while testing Vercel images

PR #780, opened 2023-08-21
Fixing scroll performance issues on certain browsers

### LINKED ISSUE #736: Extremely poor performance on Firefox Android
Landing page renders at like 10fps in Firefox on Android, which is pretty funny while reading performance claims.
Issue occurs in Firefox, but not in Chromium.

I verified two devices and two different Firefox versions, one of them is the latest nightly, one is an old stable.

I think this is because of the CSS, there are a lot of "fancy" stuff going on there, perhaps it is blurs or transitions or shadows slowing things down, but that is just a guess.

Specifically those pages are very slow:
- https://astro.build/
- https://astro.build/about
- https://astro.build/themes
- https://astro.build/showcase
- https://astro.build/blog/astro-021-preview

Those are okayish, but still laggy:
- https://astro.build/blog

Those are ok:
- https://docs.astro.build/

Landing page on a desktop Firefox too feels sluggish while scrolling, although it is pretty subtle and isn't as obvious.

tony-sull:
closes #736 

WIP: still testing the fixes here in deployed builds to confirm this fixes the issues seen in Firefox on Android

We had two big hitters on performance:
- multiple noise layers overlaid on the entire page, card backgrounds, background gradient blobs, etc
- our background grid texture was rendering a repeating linear gradient with a `mask-image` to fade it out at the edge

Unrelated to perf, this also updates our white button variant to remove the gradient text.  This is well supported these days but a detail we had talked about removing at some point for legibility and browser compatibility for any that still don't support text clipping


bholmesdev (approved):
NWTWWHB! Tested on firefox and noticed modest improvements over prod (though it isn't perfect). Not sure I can ask for much more on firefox though

PR #792, opened 2023-08-28
chore: switch out directories in `tsconfig`

ElianCodes:
I know that this isn't the biggest / important change, but it's nice to keep everything consistent


tony-sull (approved):
Ah yep, another great catch!

PR #821, opened 2023-09-14
Remove orphaned pages: careers, about

FredKSchott:
- Remove orpaned pages that are no longer linked to anywhere.
- We can always add these back later when they're needed.


ElianCodes (approved):
Ah! This is a good catch! Thanks

PR #828, opened 2023-09-20
Fix ahref issues - page has links to redirect

TheOtterlord:
Add a trailing `/` to links across the site, eliminating redirects.

I'm unsure the changes from `/chat` to `/chat/` will fix the redirect ahref issue for that, since it still redirects to a discord link, but it eliminates the redirect chain.


delucis (approved):
LGTM! Thanks for going through that so thoroughly @TheOtterlord 💜

PR #838, opened 2023-10-06
Feat: Add pricing filter to Theme Showcase

surjithctly:
Added a much-needed pricing filter to the theme showcase. So users can now filter only by free themes or paid themes depending on their needs.


natemoo-re (approved):
This seems great to me! Thanks for the PR!

tony-sull (approved):
Sorry for the delay getting this reviewed, thanks for adding in this new filter!

PR #843, opened 2023-10-11
Automatically callout the latest release post

natemoo-re:
Whether we want to do this or not is up for debate, but here's an implementation that automatically features the most recent release post on our homepage!


doodlemarks:
the people's champ, back at it again! <3

PR #851, opened 2023-10-17
fix: typo on readme for integration page

xstevenyung:
hello,

just fixing a small typo i stumble onto while browsing the code


ElianCodes (approved):
Good catch!

PR #852, opened 2023-10-19
Fix logos on brand to be correct versions

doodlemarks:
Somehow some of our assets got wonky. Also, added slightly higher res pngs and for full logo versions I added some padding so the vertical alignment is correct.


itsMapleLeaf (approved):
lgtm!

PR #854, opened 2023-10-24
feat: Add docs redirect

evadecker:
https://astro.build/docs gave a sad and cryptic 404:

image

Now it'll redirect users to https://docs.astro.build ⏩


delucis (approved):
It works! 🎉 

Thanks for noticing and adding this @evadecker 💜

PR #858, opened 2023-10-30
Add note about outdated `--experimental-static-build`

natemoo-re:
I've seen quite a few people asking about this blog post! It has fairly good SEO so people land here often. This feature flag was needed for a version of Astro prior to v1.0.0.

We might want to think of a way to automate this, but a manual `Note` inclusion seems fine for now.


tony-sull (approved):
oh yeah this would be a great one to automate, not quite sure how we'd track it but it's definitely easy to miss old blog posts when features are changed or deprecated

PR #871, opened 2023-11-12
Fix i18n Function Name for Astro 3.5 Blog Post

maxdemaio:
In the latest blog post, https://astro.build/blog/astro-350/#i18n-routing-experimental, the function name `getLocaleRelativeUrl` should be changed to `getRelativeLocaleUrl`. Using the wrong function name will cause an error when trying access the i18n pages 👍🏻 

This can be demonstrated at the following minimal reproduction repo: https://github.com/maxdemaio/i18n-repro-bug. This change also matches the docs here: https://docs.astro.build/en/guides/internationalization/#getrelativelocaleurl.

Changing this function name will solve the error and allow users to access their i18n pages! Thought it'd be a nice add for people reading the blog post.

Before:
image

After:
image


natemoo-re (approved):
Thanks for catching this!

ematipico (approved):

PR #875, opened 2023-11-14
fixes ev0 theme broken demo link

nice:
Original theme developer has changed the demo url on the repo.
https://github.com/gndx/ev0-astro-theme

Current demo link is broken.


tony-sull (approved):
Thanks for updating the link in our catalog!

PR #876, opened 2023-11-15
Add API endpoint for integrations

natemoo-re:
Resolves PLT-1159

- Adds [`/api/v1/integrations/[page]`](https://astro-build-git-feat-integration-api-astrodotbuild.vercel.app/api/v1/integrations) endpoint to fetch the data in our integration catalogue
  - Supports `search` via `?search=query`
  - Supports pagination via `/[page-number]` param and `?limit=25`
  - Supports categories via `?categories[]=featured`, `?categories[]=recent`
- Refactored existing [`/integrations`](https://astro-build-git-feat-integration-api-astrodotbuild.vercel.app/integrations) page to use the new shared utils
- The goal is to wire this up to the Astro Dev Overlay
- Adds a new step to the `update-integrations` script. Now it writes to `src/data/last-modified.json` so we can send a `Last-Modified` header along with the API responses. Considered generating an etag, but that seemed like overkill.
- ~~Currently handles `Accept: text/html` and `Accept: application/json` but I'm not sure if the HTML output makes sense?~~ Only handles `Accept: application/json` now


matthewp:
Wow, a partial 🙂

src/pages/api/v1/integrations/[...page].astro
```diff
@@ -0,0 +1,130 @@
+---
+import type { CollectionEntry } from "astro:content"
+import { integrations as LAST_MODIFIED } from "~/data/last-modified.json"
+import { getFilteredIntegrations, validateCategories } from "~/helpers/integrations.ts"
+import { paginate } from "~/helpers/paginate.ts"
+import IntegrationCard from "~/pages/integrations/_components/IntegrationCard.astro"
+import "~/styles/tailwind.css"
+
+export const prerender = false
+export const partial = true
```
Comment (natemoo-re) [inline src/pages/api/v1/integrations/[...page].astro:10]: I don't think this actually works at the moment, to be clear! We'd need to upgrade to `astro@3.4.x`

matthewp (approved):

src/pages/api/v1/integrations/[...page].astro
Comment (natemoo-re) [reply src/pages/api/v1/integrations/[...page].astro]: Resolved since this file has been removed!

tony-sull (approved):
Looks good to me!

PR #877, opened 2023-11-16
Fix CORS on integrations endpoint

natemoo-re:
Famous last words...


tony-sull (approved):
Fixing CORs in only one try 🪄

PR #878, opened 2023-11-16
Add `overlay` category to integrations

natemoo-re:
- Adds a new `overlay` category to the integrations.
- Adds a new `/api/v1/dev-overlay` endpoint. Tried to put both under the existing integrations endpoint but the router was not pleased.
- Unlike `/api/v1/integrations`, the `/dev-overlay` endpoint doesn't accept `search` or `categories[]` params (can always change this in the future)
- Adds a new property to integrations called "overlay". Pretty much the same as "featured".
- Adds a new category of integrations called "devtools". These are packages that have keywords for "devtools" or "dev-overlay".
- Does not update the Integrations page to display "overlay". It's an invisible category only for internal use.

## TODO

- [x] Migrate existing integrations to ensure they have the correct `overlay` property


natemoo-re:
Oh no. I think the problem is `trailingSlash` behavior...

tony-sull (approved):
who says Astro isn't useful for apps?

FredKSchott:
@tony-sull FYI committed a different list of 5 directly to the PR

scripts/integrations.json
```diff
@@ -39,6 +42,13 @@
     "@astropub/icons",
     "astro-json-element"
   ],
+  "overlay": [
+    "@astrojs/tailwind",
+    "@astrojs/sitemap",
+    "@astrojs/mdx",
+    "astro-compress",
```
natemoo-re, line 49: This should be removed as soon as Spotlight is available

tony-sull: what is this Spotlight magic you speak of?

PR #885, opened 2023-11-20
Update overlay endpoint

natemoo-re:
Overlay endpoint was returning an empty array. I fixed up the logic!


natemoo-re:
Preview looks good so I'm going to YOLO merge

PR #888, opened 2023-11-26
update Astro Big Doc to latest version

wassfila:
This is an update to the latest version, mainly

- the short description has been rewritten to briefly reflect the content of the repo and the value it represents.
- The image is updated to reflect the features of image panzoom and 3d viewer as an astronaut to better fit the environment
- Both webp images sizes updated and kept very low size with sufficient quality
- dropped mdx as tool, given that now this theme is based on rendering pure Markdown .md files
- shifted the deployment to Netlify due to simpler deployment without need for base_url


tony-sull (approved):
Updates look great, thanks!

PR #898, opened 2023-12-01
Fix typo in whats-new-november-2023.mdx

jasikpark:


ElianCodes (approved):
Ah! Good catch! Thanks @jasikpark

PR #902, opened 2023-12-04
new hero!

FredKSchott:
- New hero to match the messaging work on https://docs.astro.build/en/concepts/why-astro/
- new subtitle: I think this is more effective than what we have now, but it also leans back into performance more than I meant to. Might still be NWTWWHB to merge anyway... not sure how I feel about it yet tbh!
- Rebalance heading sizes to account for the new, longer/larger hero text

### DIFF
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -165,15 +165,15 @@ export default {
 				},
 
 				".heading-1": {
-					"@apply font-obviously text-7xl leading-tight": {},
-				},
-
-				".heading-2": {
 					"@apply font-obviously text-6xl leading-tight": {},
 				},
 
+				".heading-2": {
+					"@apply font-obviously text-5xl leading-tight": {},
+				},
+
 				".heading-3": {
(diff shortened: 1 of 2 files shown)


matthewp (approved):
lgtm

tailwind.config.ts
```diff
@@ -165,15 +165,15 @@ export default {
 				},
 
 				".heading-1": {
-					"@apply font-obviously text-7xl leading-tight": {},
```
tony-sull, line 168: Looks like these styles aren't used in the Hero at all, it uses the `text-` utilities directly

All good if this was an intentional change as well, but it will scale down titles on all other pages slightly 👍

FredKSchott: Because the hero is smaller, heading-3 needs to be smaller to balance that out, and then because heading-3 needs to be smaller all of these need to be re-balanced as well. There might be a way around it but after testing this on every page and a bunch of blog posts I think it looks good / is safe to ship! I kind of prefer this stylistically, as well.

PR #913, opened 2023-12-06
chore: optimize images

ElianCodes:
JPEG images were up to 4MB of data served. Ran images through sqoosh to reduce page weight


delucis (approved):
LGTM! Could probably even go smaller given these are never shown too wide, but huge improvement. Thanks for spotting @ElianCodes 🙌

FredKSchott:
Thanks @ElianCodes! I'd assumed these were all running through our image optimization, but guess not!

PR #922, opened 2023-12-11
astro 4

itsMapleLeaf:
Upgrades packages and and migrates away from the old prefetch package


astro.config.mjs
```diff
@@ -24,8 +22,10 @@ export default defineConfig({
 		solidJs(),
 		mdx(),
 		sitemap(),
-		prefetch(),
 	],
+	prefetch: {
+		defaultStrategy: "hover",
+	},
```
ElianCodes, line 28: If you wanted, you could remove this entirely, since `hover` is the default behaviour anyways

```suggestion
```

ElianCodes (approved):
Looking good and feeling smooth!

PR #924, opened 2023-12-14
chore: remove old themes

ElianCodes:
All of the removed themes result in 404's


tony-sull (approved):
Excellent, thanks @ElianCodes!

PR #945, opened 2024-01-03
Update fonts.css

jacobdalamb:
I'm not sure if this file is needed but these fonts aren't being used.

### DIFF
--- a/src/styles/fonts.css
+++ b/src/styles/fonts.css
@@ -1,36 +1,3 @@
-@font-face {
-	font-family: "MDIO";
-	src: url("/fonts/MDIO/Web/MDIO0.6-Regular.woff2") format("woff2");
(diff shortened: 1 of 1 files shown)


itsMapleLeaf:
Hey there, thanks for looking out! It's not obvious(ly, heh), but these are actually set via the tailwind config: https://github.com/withastro/astro.build/blob/35f7d90db431485aef35a4d74bf30875fc3422b7/tailwind.config.ts#L25-L27

And used via `font-mono` and `font-obviously[-wide]` classes:

https://github.com/search?q=repo%3Awithastro%2Fastro.build+font-mono&type=code
https://github.com/search?q=repo%3Awithastro%2Fastro.build+font-obviously&type=code

With that, I'll go ahead and close this

PR #948, opened 2024-01-07
Make mobile nav resources drop-down animated

onsclom:
[astro.build](astro.build) has some great animated collapsable elements already.

Screenshot 2024-01-07 at 1 52 57 PM

Using the site on mobile, I noticed the mobile nav could reuse this animation too!

## Before

https://github.com/withastro/astro.build/assets/8485687/7e287b4c-f044-4297-87f3-f0f1cded2b28

_(Notice there is also FOUC on the first open. Seems like the font loads in too late?)_

## After

https://github.com/withastro/astro.build/assets/8485687/06c2693d-7a23-4dcb-ab6e-29e4a6a22f86

## Future Considerations

The desktop version of the site works great with JavaScript disabled. However, the existing mobile nav hamburger button does not work with JavaScript disabled. If we want the mobile site to work well without JavaScript, this may need to be slightly modified too.

The existing `Collapse` component does not currently respect `prefers-reduced-motion`. I imagine the `Collapse` component can be changed to support this in a future MR.

The sidebar drop downs in Docusaurus and Nextra docs have nice animations like this, but Astro Starlight does not. Maybe this effect could be reused in Astro Starlight in some nice way?

### DIFF
--- a/src/components/HeaderDrawerContent.astro
+++ b/src/components/HeaderDrawerContent.astro
@@ -1,9 +1,6 @@
 ---
 import siteInfo from "~/data/site-info.js"
-import ChevronIcon from "~/icons/ChevronIcon.jsx"
-import LayersIcon from "~/icons/LayersIcon.jsx"
-import OpenBookIcon from "~/icons/OpenBookIcon.jsx"
-import PaletteIcon from "~/icons/PaletteIcon.jsx"
+import ResourcesDropdown from "./ResourcesDropdown.tsx"
 import SocialLinks from "./SocialLinks.astro"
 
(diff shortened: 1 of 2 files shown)


itsMapleLeaf:
Hey, sorry for the delay on looking at this! I opened up the preview, and it looks like you can't click any links when it's open:

https://github.com/withastro/astro.build/assets/19603573/f724b0a7-9570-4882-b3d1-f090f6b058d2

itsMapleLeaf:
Also: it's not a problem if it's not animated without JS, just as long as it opens ¯\\\_(ツ)\_/¯

itsMapleLeaf:
This has gotten pretty stale, so I'm gonna close this

PR #960, opened 2024-01-22
Fix prerender doc link

tunetheweb:
Noticed a typo in the post


delucis (approved):
Nice catch and thanks for fixing! 🚀

PR #983, opened 2024-02-06
Update press assets

delucis:
- I tweaked the description of the monochrome logo on dark card to not be identical to gradient variant description.
- Added light/dark Starlight logos.
- I considered adding a Houston, but a) not an instant export to SVG because of the conic gradient and b) I’m not 100% sure how much Houston should be used by third parties.


natemoo-re (approved):
This seems entirely reasonable to me!

public/assets/press/starlight-logo-dark.svg
```diff
@@ -0,0 +1,35 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 200 55">
```
natemoo-re, line 1: Run these through [`svgomg`](https://jakearchibald.github.io/svgomg/)?

natemoo-re (approved):

public/assets/press/starlight-logo-dark.svg
delucis: I did 😅 

There are probably a few small wins — I don’t think SVGO is able to deduplicate identical gradient filters — but not much.

PR #996, opened 2024-03-03
fix: Adjustment of card styles

ktym4a:
The layout was strange when the content was small, and this fix, it always fills the grid width

like this page:
https://astro.build/integrations/2/?categories%5B%5D=recent

## before
Screenshot 2024-03-03 at 23 29 26

## after
Screenshot 2024-03-03 at 23 29 12


ktym4a:
I don't think it should be included in this PR, but I think we also need to consider when there is something in `homepageUrl` that is not a URL.

like `astro-auth`.
because the link for that is `todo:` now.

PR #1000, opened 2024-03-06
Add `/db/*` redirects

natemoo-re:
We have an existing `astro.build/config` redirect. Would be nice to have the same for DB stuff!

These are being used in https://github.com/withastro/astro/pull/10349 but don't exist yet.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -90,11 +90,11 @@
     },
     {
       "source": "/db/config/",
-      "destination": "https://docs-git-sarah-patch-6-astrodotbuild.vercel.app/guides/astro-db/#configuration"
+      "destination": "https://docs-git-sarah-patch-6-astrodotbuild.vercel.app/en/guides/astro-db/#configuration"
     },
     {
       "source": "/db/seed/",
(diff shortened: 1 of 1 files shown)


vercel.json
```diff
@@ -88,6 +88,14 @@
       "source": "/config/",
       "destination": "https://docs.astro.build/reference/configuration-reference"
     },
+    {
+      "source": "/db/config/",
+      "destination": "https://docs.astro.build/reference/db-configuration/"
```
natemoo-re, line 93:
```suggestion
      "destination": "https://docs.astro.build/guides/astro-db/#configuration"
```

vercel.json
```diff
@@ -88,6 +88,14 @@
...
+    },
+    {
+      "source": "/db/seed/",
+      "destination": "https://docs.astro.build/guides/db/#seed"
```
natemoo-re, line 97:
```suggestion
      "destination": ""https://docs.astro.build/guides/astro-db/#seed-development-data"
```

PR #1001, opened 2024-03-08
Use variable fonts instead of individual weights

natemoo-re:
Updates our font stack! 

- Rather than loading individual weights, all fonts now point to a variable cut.
- Matches the existing individual weights with variable font settings.
- Removes `@fontsource/inter` because Inter is now served by our own font CDN.
- Should improve loading performance, but that needs to be measured.


itsMapleLeaf (approved):

FredKSchott:
image

a bit shocked to see the homepage so clearly broken by this, and judging by the commit history on main this wasn't a smooth rollout either. Let's be much more careful merging major changes to our public marketing site like this, and definitely not on a Friday at 6pm CT.

PR #1014, opened 2024-03-13
Fix broken code example

morinokami:
Hi, I found a small mistake in the auth code example and fixed it.


ElianCodes (approved):
Thanks for catching!

PR #1028, opened 2024-03-16
Add "slnt" 0 for `font-obviously` to prevent unexpected italic in safari.

ktym4a:
about [message from discord](https://discord.com/channels/830184174198718474/872579324446928896/1218555975234097214).

I think this will fix it, but I would like to check preview once.


OliverSpeir (approved):
Approving to see deployment

edit: I can't approve to see deployment 😅

delucis:
Authorized the preview deploy — should be here shortly. Thanks for tackling @ktym4a!

ktym4a:
Looks good to me in Safari on MacOS!
and I don't have IOS, so I can't check about that.

natemoo-re:
Thank you _so_ much for the PR! You got this fixed way before we did and we should have merged it right away.

It took a bit longer, but I reworked our font handling completely in https://github.com/withastro/astro.build/pull/1031. We are now subsetting our font and have eliminated the `slnt` axis for performance reasons!

PR #1031, opened 2024-03-19
Fix font performance regression

natemoo-re:
Whoops, we tanked our performance by merging https://github.com/withastro/astro.build/pull/1005. 

This PR drastically improves our implementation by switching to some subset fonts


src/styles/fonts.css
```diff
@@ -38,11 +40,3 @@
...
 	line-gap-override: 19%;
 }
-
-@font-face {
-	font-family: obviously-wide-fallback;
-	src: local("Arial Black");
-	size-adjust: 134%;
-	ascent-override: 96%;
-	descent-override: 27%;
-}
```
delucis, line 48: Worth double checking this — we put it in place to reduce CLS when the heavier weight Obviously loaded in vs system fonts.

package.json
```diff
@@ -66,6 +66,7 @@
     "eslint-plugin-jsx-a11y": "^6.8.0",
     "eslint-plugin-prettier": "^5.0.1",
     "eslint-plugin-tailwindcss": "^3.13.0",
+    "glyphhanger": "^5.0.0",
```
delucis, line 69: Guessing this ended up not getting used in this repo and should be uninstalled?

delucis (commented):
I know you said this still needs wrapping up, so sorry for the early review, but had a moment to look at this. Feel free to ignore comments where stuff is WIP. Thanks so much for jumping on this Nate!

PR #1033, opened 2024-03-22
Add support for transparent nav bar

natemoo-re:
- Adds a transparent style navbar
- Updates the `/db` page to use it
- Updates the `/db` page to improve hero image loading experience

### DIFF
--- a/src/helpers/disclosure.ts
+++ b/src/helpers/disclosure.ts
@@ -1,4 +1,5 @@
 import { createEffect, createSignal } from "solid-js";
+import { EventEmitter } from './emitter.ts';
 
 export function createDisclosure({
 	button,
@@ -25,10 +26,11 @@ export function createDisclosure({
 		});
 	}
 
-	const [visible, setVisible] = createSignal(false);
+	const emitter = new EventEmitter<{ toggle: { visible: boolean } }>();
+	const [visible, setVisible] = createSignal(false)
 
 	createEffect(() => {
-		button.setAttribute("aria-expanded", String(visible()));
+		button.setAttribute("aria-expanded", String(visible()))
 
 		if (visible()) {
 			content.style.removeProperty("display");
@@ -104,4 +106,6 @@ export function createDisclosure({
 			window.removeEventListener("focusin", handleFocusLost);
 		}
 	});
+
+	return emitter;
 }
--- a/src/helpers/disclosure.ts
+++ b/src/helpers/disclosure.ts
@@ -95,6 +95,7 @@ export function createDisclosure({
 	};
 
 	createEffect(() => {
+		emitter.emit('toggle', { visible: visible() });
 		if (visible()) {
 			window.addEventListener("click", handleClickOutside);
 			window.addEventListener("focusin", handleFocusLost);
--- a/src/layouts/MainLayout.astro
+++ b/src/layouts/MainLayout.astro
@@ -1,24 +1,25 @@
 ---
 import BaseHead, { type Props as HeadProps } from "~/components/BaseHead.astro"
 import Footer from "~/components/Footer.astro"
-import Header from "~/components/Header.astro"
+import Header, { type Props as HeaderProps } from "~/components/Header.astro"
 import HeaderDrawerContent from "~/components/HeaderDrawerContent.astro"
 import HeaderNav from "~/components/HeaderNav.astro"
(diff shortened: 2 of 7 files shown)


src/helpers/disclosure.ts
```diff
@@ -26,9 +26,17 @@ export function createDisclosure({
...
+	const footer = document.querySelector("footer")
 
 	createEffect(() => {
 		button.setAttribute("aria-expanded", String(visible()))
+		if (main) {
+			main.inert = visible()
+		}
+		if (footer) {
+			footer.inert = visible()
+		}
```
itsMapleLeaf, line 39: Doing an if check on these is a silent failure if the elements happen to change 🫠

I'd consider something like this:
```ts
	const main = expect(document.querySelector("main"))
	const footer = expect(document.querySelector("footer"))
```
```ts
// in some helpers file
export function expect<T>(value: T): NonNullable<T> {
	if (value == null) {
		throw new Error(`Expected a non-nullable value, got ${value}`)
	}
	return value
}
```

src/layouts/MainLayout.astro
```diff
@@ -1,24 +1,25 @@
...
 ---
 
 <!doctype html>
 <html
 	lang="en"
 	class:list={[
-		"astro-gray-700 overflow-x-hidden break-words bg-black text-astro-gray-100 [color-scheme:dark] [word-break:break-word]",
-		classList,
+		"astro-gray-700 overflow-x-hidden break-words text-astro-gray-100 [color-scheme:dark] [word-break:break-word]",
+		classList ? classList : "bg-black",
```
itsMapleLeaf, line 22:
```suggestion
		classList || "bg-black",
```

src/layouts/MainLayout.astro
```diff
@@ -47,6 +55,23 @@ const { class: classList, ...head } = Astro.props
...
+					left: 0;
+					z-index: 11;
+					background: #0d0f14;
+					pointer-events: none;
+					opacity: 0;
+					transition: opacity 400ms cubic-bezier(0.23, 1, 0.32, 1);
+				}
+				:root.disclosure-open body::before {
+					opacity: 1;
+				}
```
itsMapleLeaf, line 74: This is tailwindable, but I won't block on that ¯\\\_(ツ)\_/¯

src/pages/db/index.astro
```diff
@@ -32,7 +33,7 @@ const description = "The SQL database platform for content-driven websites."
 	</div>
 </MainLayout>
 
-<style>
+<style is:global>
 	.db {
 		background-color: #060913 !important;
 	}
```
itsMapleLeaf, line 39: I'd definitely rather pass this as a dynamic class name than make it global and `!important` 😂

itsMapleLeaf (approved):
I'll stamp this with some optional nits

src/helpers/disclosure.ts
itsMapleLeaf: Also... it seems kind of scary for this to operate on global elements? It was created to mostly be a self-contained abstraction 🤔 

I also just realized that this is written in a way that `main` and `footer` might not exist on the page, so ignore the first comment I suppose! 🤪

natemoo-re: Ahh right, that's a good catch. This file is in the `src/helpers` folder, but the only place it is used is for the header... I agree that this global stuff shouldn't go here, though. I'll come up with something that makes more sense.

src/pages/db/index.astro
natemoo-re: I swear there was a good reason to have the style block here at one point, I just totally forgot to remove it now that there's only a single style 😂

src/layouts/MainLayout.astro
natemoo-re: I anticipate that we will be having this conversation frequently 😂

Pseudo-elements feel extra strange in Tailwind, but tbh all of Tailwind feels pretty weird to me. I'll just need some time to adjust my brain.

src/helpers/disclosure.ts
natemoo-re: Updated with an EventEmitter pattern in [394e38e](https://github.com/withastro/astro.build/pull/1033/commits/394e38ec4480a478dd3b6e54b0390452e8677e9d)

PR #1050, opened 2024-04-05
Updates use cases markup and styling

iann-mathaiya:


src/pages/db/_components/use-cases/BlogFilters.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+			{filteredPosts().length > 0 ? 
+        <ul id="postsList" class="space-y-4 divide-y divide-astro-gray-600">
+				{filteredPosts().map((post) => (
+						<li class="space-y-2 pt-4">
+							<h4 class="text-lg text-[#9FA5B3]">{post.title}</h4>
+							<p class="text-balance text-sm text-[#6E7584]">{post.summary}</p>
+							<p class="text-sm text-astro-gray-400">{post.publishedOn}</p>
+						</li>
+				))}
```
itsMapleLeaf, line 65: Although this _works_, should prefer using the `<For>` component in solid

src/pages/db/_components/use-cases/Reactions.tsx
```diff
@@ -0,0 +1,43 @@
...
+				return { ...reaction, count: reaction.count + 1, isClicked: true }
+			}
+			return reaction
+		})
+		setReactions(updatedReactions)
+	}
+
+	return (
+		<div class="mt-4 flex items-center justify-start gap-4 text-astro-gray-300">
+			{reactions().map((reaction) => (
```
itsMapleLeaf, line 31: Same here, prefer using `<For>`

src/pages/db/_components/use-cases/Reactions.tsx
```diff
@@ -0,0 +1,43 @@
...
+		{ id: 2, emoji: "👍", count: 459, isClicked: false },
+		{ id: 3, emoji: "👎", count: 391, isClicked: false },
+		{ id: 4, emoji: "🚀", count: 275, isClicked: false },
+		{ id: 5, emoji: "🎉", count: 137, isClicked: false }
+	])
+
+	function handleReactionClick(id: number) {
+		const updatedReactions = reactions().map((reaction) => {
+			if(selected() === true && reaction.id === id){
+				setSelected(false)
```
itsMapleLeaf, line 16: Calling state setters (or doing any side effects) in a `.map` callback is potentially confusing; `.map` callbacks should ideally be pure.

itsMapleLeaf (changes requested):
Hi, thanks for the contribution! I'd like to see some things addressed before merging:

- The images here are broken on deployment:
	image

- Use `.webp` format for the images so they aren't as large

- The form elements on the "Comments" section and "Forms" sections shouldn't actually make form submissions; use `type="button"` for those instead of `type="submit"`

- On the code for the "Blog" example, there's a missing trailing comma on the `title:` line

- The "Sign Up" link on the "Auth" example should use a button element instead of a link element, since it doesn't actually lead anywhere

- The functionality on these reaction buttons doesn't feel right

https://github.com/withastro/astro.build/assets/19603573/e3248972-5a90-4e28-a30d-f0f670dc8213

natemoo-re (commented):
Mostly looks good! Thanks Ian.

I'm not sure if we need to use Solid for the interactive parts on this page. I'm not opposed to it, but it doesn't seem all that necessary to me. Can we keep these to static `.astro` components for now? I'm going to work on wiring everything up later.

iann-mathaiya:
> Mostly looks good! Thanks Ian.
> 
> I'm not sure if we need to use Solid for the interactive parts on this page. I'm not opposed to it, but it doesn't seem all that necessary to me. Can we keep these to static `.astro` components for now? I'm going to work on wiring everything up later.

I've updated the interactive parts to vanilla js. I've also tackled the issues @itsMapleLeaf had raised but I can't test if the images are rendering as intended on deployment since I don't have authorization to deploy

Screenshot 2024-04-08 at 13 19 16

src/pages/db/_components/use-cases/Feedback.astro
```diff
@@ -1,47 +1,205 @@
...
+      celebrationReactionButton.innerText = `🚀 ${celebration}`
+      celebrationReactionButton.removeAttribute("aria-selected")
+    }
+  }
+
+  fireReactionButton.addEventListener("click", toggleFire)
+  thumbsUpReactionButton.addEventListener("click", toggleThumbsUp)
+  thumbsDownReactionButton.addEventListener("click", toggleThumbsDown)
+  celebrationReactionButton.addEventListener("click", toggleCelebration)
+  rocketLaunchReactionButton.addEventListener("click", toggleRocketLaunch)
```
itsMapleLeaf, line 204: Sorry, but this doesn't look very fun to maintain, and seems prone to breakage 😅 

I would wrap the functionality up in a function:
```js
function createToggleButton(count: number, emoji: string, button: HTMLButtonElement) {
	let enabled = false;
	button.addEventListener("click", () => {
		enabled = !enabled;
		const shownCount = count + (enabled ? 1 : 0);
		button.innerText = `${emoji} ${shownCount}`;
		button.setAttribute("aria-selected", String(enabled));
	});
}
```

I definitely appreciate the initiative to reduce bundle size, however!

src/pages/db/_components/use-cases/Comments.astro
```diff
@@ -1,57 +1,100 @@
...
 import code from "./snippets/comments.txt?raw"
+import { Code } from "astro-expressive-code/components"
+
+import wadePic from "../../_assets/wade-warren.png"
+import jacobPic from "../../_assets/jacob-jones.png"
+import marvinPic from "../../_assets/marvin-mckinney.png"
+
+const wadeProfilePicture = await getImage({ src: wadePic, format: "webp", width: 40 })
+const marvinProfilePicture = await getImage({ src: marvinPic, format: "webp", width: 40 })
+const jacobProfilePicture = await getImage({ src: jacobPic, format: "webp", width: 40 })
```
itsMapleLeaf, line 14: I did not notice the webp conversion here! Nice!

itsMapleLeaf (changes requested):
Also make sure you actually delete the `package-lock.json` file and commit. Just adding it to the `.gitignore` won't remove it from the PR.

src/pages/db/_components/use-cases/Feedback.astro
iann-mathaiya: Yeah, this is cleaner and easier to maintain

itsMapleLeaf (approved):
Looks and feels fantastic. Thanks again @iann-mathaiya!

PR #1065, opened 2024-04-12
fix: make DB fetched types typesafe

ElianCodes:
Adds the DB types as actual types

### DIFF
--- a/src/components/ImageGallery.astro
+++ b/src/components/ImageGallery.astro
@@ -1,6 +1,5 @@
 ---
 import type { HTMLAttributes } from "astro/types"
-import { Image } from "astro:assets"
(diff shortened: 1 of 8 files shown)


itsMapleLeaf (approved):
I'll approve since this is a huge improvement over what we have (NWTWWHB!) but I'll note that "typesafe" doesn't just mean "types are present". It can actually often mean the opposite: that the types that were added are lies! 😱 This can silently blow up if the API changes and we forget to update the codebase (`[object Object]` and `NaN` anyone?)

The _real_ path to type safety is proving at runtime that a certain unknown value is what you need it to be. That can be as simple as `typeof value === 'string'`, or going to the other end of the spectrum with a validation lib

I'd definitely prefer to pick up zod at some point, but this unblocks the biome update, so I'm keen to have this in 😄

PR #1067, opened 2024-04-12
feat: enable ISR

ElianCodes:


ElianCodes:
* **#1067** <a href="https://app.graphite.dev/github/pr/withastro/astro.build/1067?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @ElianCodes and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

lilnasy:
ISR cache wouldn't look at search params, so https://github.com/withastro/astro.build/blob/main/src/pages/integrations/%5B...page%5D.astro (as one example) wouldn't work out of the box.

natemoo-re:
@lilnasy in that case, what would you recommend using here? A [`cdn-cache-control`](https://vercel.com/docs/edge-network/headers#cdn-cache-control-header) header?

lilnasy:
Yeah, standard cache control would be my first attempt before vendor-specific ISR.

ISR gives you persistent and global cache. We could see if the header's ephemeral and per-region cache works for us first.

PR #1073, opened 2024-04-15
update biome (for real this time)

itsMapleLeaf:
I split this into two commits, one for the biome updates and the other for the formatting.


ElianCodes (approved):
Yay

PR #1059, opened 2024-04-09
[DB] animated feature cards

natemoo-re:
Updates the bento box features section of the `/db` page

PR #1079, opened 2024-04-26
Improve DB intersection observer logic

natemoo-re:
The `IntersectionObserver` logic added to `/db` in #1059 consistently skipped highlighting the **Features** section. 

This PR refactors the logic to track the delta between ratios rather than using a naive `entry.isIntersecting` check.

Tiny perf improvement by hoisting the `selector` variable up, too.

### DIFF
--- a/src/pages/db/_components/Subnav.astro
+++ b/src/pages/db/_components/Subnav.astro
@@ -68,22 +68,18 @@ import Badge from "./Badge.astro";
 		.map((a) => a.getAttribute("href")?.slice(1))
 		.filter(Boolean);
 	const selector = `:is(${targets.map((id) => `[id="${id}"]`).join(",")})`;
-	const ids: Record<string, number> = {};
+	const ids = new Map<string, number>();
 	const io = new IntersectionObserver(
 		([entry]) => {
-			console.log(
-				entry.target.querySelector(selector),
-				entry.intersectionRatio,
-			);
 			const element = entry.target.querySelector(selector);
 			if (!element) return;
-			const prevRatio = ids[element.id] ?? 0;
+			const prevRatio = ids.get(element.id) ?? 0;
 			if (entry.intersectionRatio > prevRatio) {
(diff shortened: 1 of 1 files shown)


src/pages/db/_components/Subnav.astro
```diff
@@ -67,18 +67,23 @@ import Badge from "./Badge.astro";
 	const targets = Array.from(nav.querySelectorAll("a[href^='#']"))
 		.map((a) => a.getAttribute("href")?.slice(1))
 		.filter(Boolean);
+	const selector = `:is(${targets.map((id) => `[id="${id}"]`).join(",")})`;
+	const ids: Record<string, number> = {};
```
itsMapleLeaf, line 71: I'd have used a `Map` here, but no big deal

src/pages/db/_components/Subnav.astro
```diff
@@ -67,18 +67,23 @@ import Badge from "./Badge.astro";
...
-				);
-				if (element) {
-					const link = nav.querySelector(`a[href="#${element.id}"]`);
-					resetSubnav();
-					link?.setAttribute("aria-current", "true");
-				}
+			console.log(
+				entry.target.querySelector(selector),
+				entry.intersectionRatio,
+			);
```
itsMapleLeaf, line 77: Consider removing this log?

src/pages/db/_components/Subnav.astro
```diff
@@ -67,18 +67,23 @@ import Badge from "./Badge.astro";
...
+			const element = entry.target.querySelector(selector);
+			if (!element) return;
+			const prevRatio = ids[element.id];
```
itsMapleLeaf, line 80: It looks like this could be undefined. JS being JS it probably still works, but I'd prefer a more explicit comparison

```suggestion
			const prevRatio = ids[element.id] ?? 0;
```

itsMapleLeaf (approved):
Approved with suggestions

PR #1089, opened 2024-05-07
Adjust main CTA link

natemoo-re:
Rather than linking the main CTA to https://docs.astro.build (which is another landing page), we should link directly to https://docs.astro.build/en/install/auto/. 

This PR implements that change.


itsMapleLeaf (approved):
LGTM!

PR #1093, opened 2024-05-08
Correct username that recieved a community award

sasoria:
It looks like my username is wrong in the list of Community awards from march. This pull request updates it to the correct handle (@sasoria).


sarah11918 (approved):
So sorry about that @sasoria ! When we write these, we're often using Discord usernames, so we'll double check that they are GitHub ones in future!

PR #1102, opened 2024-05-17
chore: update showcase

evadecker:
My preview on the showcase was a bit out of date so I generated some new files and updated to my new domain name!


delucis (approved):
Thanks for the update @evadecker!

PR #1118, opened 2024-06-06
Update @sentryastro.md

smeubank:
On Astro build there is a link to set up an integration with Sentry, but it doesn't take the developer to the most useful page.

https://astro.build/integrations/?search=sentry

Should take the developer here, for the latest info on the Sentry Astro SDK


FredKSchott:
Thanks @smeubank!

PR #1110, opened 2024-05-27
Homepage revamp

iann-mathaiya:

PR #1120, opened 2024-06-10
Fix Pill Link to Astro v4.10 release blog post

djsiddz:
The Pill currently links to the previous content's resource.

Users would expect to read about the 4.10 release instead of being redirected to Astro DB on clicking on the Pill.


florian-lefebvre:
Hey thanks for the PR! This is going to be addressed in a huge rework already (#1110) so I'm going to close it

PR #1124, opened 2024-06-11
Fixing Footer Link for Terms and Conditions Link broken

MostlyWhat:
As title, it keeps appending terms when clicking to the end of the url.


FredKSchott:
Thanks!

PR #1129, opened 2024-06-12
Implement alternative background for Firefox

TheOtterlord:
This adds a PNG background specifically for Firefox, fixing both the performance and rendering issues Firefox has with the SVG


TheOtterlord:
Blocked. Looks like something changed since I tested yesterday, and now the image is off-centre :thinking:

PR #1130, opened 2024-06-12
Reduce Hero logo sizes

TheOtterlord:
Reduces the size of each team logo to 2x the specified height of each image for 2x DPR. I'm unsure if there are any/many displays of higher DPR, but we could increase this if needed.


delucis (commented):
Any reason these aren’t SVGs? At least on my screen Microsoft and NBC logos look a teeny bit blurry here. I’d imagine all of these are available as SVG somewhere.

delucis (approved):
Thanks again for kicking this off @TheOtterlord — I ended up replacing these with SVGs. The reason some of the PNGs were blurry was even the original images were added at too low a resolution, so it was simpler to just track down the couple of missing SVGs as most of the logos were already available in SVG form in this repo.

- lighterweight
- inlined, so zero CLS/loading flash
- profit

While doing this, I also realised the directory full of PNG logos these were from is now completely unused, so I took the liberty to delete it entirely to encourage use of SVGs in the future.

delucis:
Build failure was just a flakey run unrelated to this PR, here’s the re-run which passed: https://app.netlify.com/sites/astro-www-2/deploys/66c3baee388c8542555260e4

PR #1143, opened 2024-06-17
Update future-of-astro-content-layer.mdx

hfournier:
Fixed word order typo changing "will that" to "that will".


sarah11918 (approved):
Thank you @hfournier ! 🙌

PR #1145, opened 2024-06-18
Astro 4.11

matthewp:


sarah11918 (commented):
Looks great @matthewp ! Left a few tiny nitpicks for your consideration!

src/content/blog/astro-4110.mdx
```diff
@@ -0,0 +1,83 @@
...
+npm install astro@latest
+pnpm upgrade astro --latest
+yarn upgrade astro --latest
+```
+
+## Improvements to 500.astro
+
+The `500.astro` is a special page in `src/pages` that allows you to handle [500](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) errors gracefully. A 500 error most commonly happens where there is a bug in your application; such as not wrapping delicate code in a `try/catch`. When component code throws, Astro cathes and returns a 500 response. If you have `src/pages/500.astro` this page gets rendered into the response body.
+
+New in 4.11, the 500.astro page now receives the `error` as a prop. This allows you to give more contextual information about the error. For example in dev mode you might show the error's stack trace to aid in debugging.
```
florian-lefebvre, line 39: I know it mentions dev mode but we're actually recommending against showing sensitive information on this page in docs, eg. the error stack trace since it can show the structure of the user's code. Maybe worth using a similar example to what we have in docs:
```astro
<div>{error instanceof Error ? error.message : 'Unknown error'}</div>
```

src/content/blog/astro-4110.mdx
```diff
@@ -0,0 +1,83 @@
...
+import BlogSummary from "/src/components/BlogSummary.astro";
+
+<BlogSummary>
+  Astro 4.11 is out with improvements to the 500.astro special page and Shiki transformers in the Code component.
+</BlogSummary>
+
+Learn more on each of these:
+
+- [**Improvements to 500.astro**](#improvements-to-500astro)
+- [**Shiki transfomers in the Code component**](#shiki-transfomers-in-the-code-component)
```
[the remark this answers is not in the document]
matthewp: I think that "Full release highlights include" reads like there's more than what is explained in the summary, but there's not, it's exactly the same. That's why I didn't go with such a description, as we usually do.

src/content/blog/astro-4110.mdx
matthewp: Is the error ever anything other than an Error? What are the other types?

florian-lefebvre: It can be anything. If someone throws a string it can be a string

matthewp: ok updated, thanks!

src/content/blog/astro-4110.mdx
```diff
@@ -0,0 +1,83 @@
...
+lang: "en"
+---
+
+import BlogSummary from "/src/components/BlogSummary.astro";
+
+<BlogSummary>
+  Astro 4.11 is out with custom 500 page improvements and Shiki transformers in the Code component.
+</BlogSummary>
+
+Full release highlights include:
```
matthewp, line 18:
```suggestion
Learn more on each of these:
```

PR #1146, opened 2024-06-19
Refresh showcase CI scraper

delucis:
This PR is mostly some housekeeping around our automated CI script that updates showcase sites.

- I manually added some sites that due to animated heroes were failing to scrape every time and slowing the CI script down
- I added some URLs to our block list for sites that were 404-ing each time we scrape
- I tweaked how we filter the sites to scrape to use local in-repo data instead of the live `showcase.json` (not super relevant for CI runs, but handy for local work where you may have changes not reflected in the prod JSON data)
- Sites built with Starlight are now automatically categorised by the scraper, so we can filter these if we need to.
- Existing Starlight sites in the catalogue have had the `'starlight'` category backfilled.
- I cleaned up the content collection schema for showcase items so that it reflects the already in-use `categories` property (this was applied to many existing files, but ignored in the schema). I caught a few typos etc. doing this too.
- Showcase sites now have a `dateAdded` field. I backfilled this with Git data and going forward this will be added automatically by the scraper for new sites.

For some context: currently working on supporting more automated showcase components for our monthly blog posts for @sarah11918, so this is preliminary work to unblock that.


sarah11918 (approved):
Checked a few blog post previews, and verifying they still look fine!

PR #1148, opened 2024-06-20
fixes broken URL in May 2024 blog post

sarah11918:
This moves a theme from the themes component to a text link because the link in the themes catalogue is broken


delucis (approved):
Thanks @sarah11918!

PR #1157, opened 2024-06-28
Update showcase content entry

jeremyfrank:
**Summary:**

This PR adjusts the `title` and `url` and two images for the explorations.viget.com entry in the showcase content collection.

**Background:**

Last fall, one of my teammates submitted some sites for inclusion in the showcase [here](https://github.com/withastro/roadmap/discussions/521#discussioncomment-7394457). Three of those sites were:

* [Viget Explorations: Mapping Design Potentials](https://explorations.viget.com/)
* [AI assisted meal selection: A Viget Exploration](https://explorations.viget.com/ai-food/)
* [Improving the Provider Finder Experience](https://explorations.viget.com/healthcare/)

The ci environment ran the update-showcase script and opened the PR [here](https://github.com/withastro/astro.build/pull/865). Of those three links, only the last one was added.

**Details:**

My hunch as to why this happened is because these three Astro sites are part of a multi-site same-domain setup, each in their own sub-directory. The `ShowcaseScraper` creates markdown files based off of the `origin`. This is 100% correct btw, to prevent the addition of sub-pages.

So in this case, a markdown file was created for each link, but w/ the same origin, overwriting the previous file. I don't think there's a good way to detect multi-site and same-domain scenarios like this in the update-showcase script, so I just chose to manually update the content entry.


delucis (approved):
Thanks for this @jeremyfrank! Yeah we filter to one per origin as it’s fairly common to share e.g. homepage + blog post about Astro, which is awesome and I’m glad they do, but then it makes sense to only include one. Appreciate you following up 🙌

PR #1158, opened 2024-06-28
Adjust timeout for showcase screenshots

jeremyfrank:
**Summary:**

Adds the [timeout option](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback#timeout) to the `requestIdleCallback` when getting screenshots for showcase sites.

**Background:**

After opening #1157, I added a site to https://github.com/withastro/roadmap/discussions/521#discussioncomment-9904855, then ran the update-showcase script locally to see things run. The script timed out on the new site, due to requestIdleCallback not resolving. This probably has something to do with animations running on load and an idle state never being reached for this particular site.

Before | After
--- | ---
image | image

### DIFF
--- a/scripts/update-showcase.mjs
+++ b/scripts/update-showcase.mjs
@@ -347,7 +347,7 @@ class ShowcaseScraper {
 		console.log("Waiting for page to settle");
 		await page.evaluate(() => {
 			return new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


delucis (approved):
Seems reasonable — thanks @jeremyfrank!

PR #1185, opened 2024-07-22
Integrations catalogue tweaks

delucis:
- Removes the concept of “featured” integrations entirely

  We were not good about keeping this updated and download count is a reasonable proxy for popularity and what gets featured.

- Adds a new `downloadFactor` field to support downranking integrations with lots of downloads. This was needed for a couple of integrations that are bundled in some more popular package but don’t actually get used as much as the npm download count suggests. For example, `@astrojs/prism` is a dependency of `astro`, so is kind of automatically the “most popular integration”. But in reality only like 0.05% of Astro users actually enable it.

- Tweaked the wording in the integrations catalogue headings from “All integrations” to “Popular integrations”. This text only ever displays on the catalogue homepage and follows the “Recently added” heading, so feels like it might be a better counterpart.

- Displays a heading when a user has searched for something. Previously, search results had no heading. Now, the heading reads “Results for "query"”.

- Updates our integration script to properly escape Markdown in integration descriptions with `mdast` utilities. Currently this is done with a regular expression that tries to strip Markdown links, but leads to stuff like this:

  astro-htmx integration card with a garbled description containing partial Markdown syntax for bold text and a link

  Tested this fix locally but going to rerun CI after this merges to apply the updated descriptions rather than doing that in this PR

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [x] Safari
- [ ] iOS Safari

### DIFF
--- a/src/content/integrations/@unocssastro.md
+++ b/src/content/integrations/@unocssastro.md
@@ -8,4 +8,6 @@ npmUrl: "https://www.npmjs.com/package/@unocss/astro"
 repoUrl: "https://github.com/unocss/unocss"
 homepageUrl: "https://github.com/unocss/unocss#readme"
 downloads: 975889
+# unocss bundles its Astro integration with all installs, real usage is much lower than the download count suggests
+downloadFactor: 0.02
 ---
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
 	"devDependencies": {
 		"@actions/core": "^1.10.1",
 		"@astrojs/check": "^0.5.6",
-		"@astrojs/site-kit": "github:withastro/site-kit",
 		"@astrojs/netlify": "^5.3.5",
+		"@astrojs/site-kit": "github:withastro/site-kit",
 		"@biomejs/biome": "1.7.0",
 		"@octokit/graphql": "^7.0.2",
 		"@types/json-to-pretty-yaml": "1.2.1",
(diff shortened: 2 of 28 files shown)


src/content/integrations/@unocssastro.md
```diff
@@ -8,4 +8,6 @@ npmUrl: "https://www.npmjs.com/package/@unocss/astro"
 repoUrl: "https://github.com/unocss/unocss"
 homepageUrl: "https://github.com/unocss/unocss#readme"
 downloads: 975889
+# unocss bundles its Astro integration with all installs, real usage is much lower than the download count suggests
+downloadFactor: 0.02
```
itsMapleLeaf, line 12: So this is just an arbitrary number to lower inflated download counts? I feel like this could open up a can of worms, but I can't think of anything better

itsMapleLeaf (approved):
approved with tiny concern

src/content/integrations/@unocssastro.md
delucis: In case it’s any reassurance, it’s not 100% arbitrary — I calculated this based on some telemetry data to scale it to the rough ballpark. But yeah, obviously a bit of a weird one. Luckily there only seemed to be a couple of these where the parent package is so popular that it caused issues.

sarah11918 (approved):
Tested on Chromebook and the new changes are great. I'll report back after I've done Android tablet and phone, but approving all the changes

PR #1189, opened 2024-07-23
Fix card overflow

TheOtterlord:
Fix an overflow in the sponsor cards on desktop.
image

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/pages/_components/landing-page/Sponsors.astro
+++ b/src/pages/_components/landing-page/Sponsors.astro
@@ -72,7 +72,7 @@ const sponsors = [
 						href={sponsor.href}
 						rel="sponsored"
 						target="_blank"
-						class="size-full basis-1/3 flex flex-col lg:grid lg:grid-rows-subgrid lg:row-span-full items-center justify-center gap-2 sm:gap-4 -outline-offset-2 hover:bg-astro-gray-100/5 transition-colors duration-400"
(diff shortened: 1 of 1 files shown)


delucis (changes requested):
Thanks @TheOtterlord! I think this is actually caused by a difference in how Chrome and Firefox handle subgrid and `gap` and we should try to fix this in Firefox properly rather than just clipping the overflow.

Might work just to move the `gap` up to the parent grid instead of setting it on the child? Let me know if you’d like me to take a closer look.

TheOtterlord:
Thanks, that worked great.

delucis (approved):
Nice, looks great now in both Chrome and Firefox to me. Thanks @TheOtterlord!

delucis:
Hadn’t checked on a small viewport but looks like this removed the gap there:

| before | after |
|---|---|
| image | image

TheOtterlord:
OK, managed to get it working in the end :sweat_smile:

Retested in Firefox/FF Android/Chromium/Chromium Android

delucis (commented):
Looks good now! Should we still remove the `sm:gap-4` on the child? Or is that still needed?

TheOtterlord:
Still needed. Removing it causes the card to not take up the full height of the parent.

image

delucis (approved):
Thanks. Probably needs something like `sm:gap-[inherit]` technically, but I think this is good to me. Works as expected and we’ve already spent enough time on it 😁 

Nice work @TheOtterlord!

PR #1191, opened 2024-07-25
Fixes broken links on Astro Integrations page 

jamielife:
Fixed issue where GitHub URL was spelled "advenced" instead of "advanced", linking to 404 page on GitHub visible here: https://astro.build/integrations/?search=toast 

- [✔] Chrome / Chromium
- [✔] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


delucis (commented):
Thanks for the PR @jamielife! The integrations catalogue pulls data directly from npm automatically, so even if we fix this, it will break again, because the URL is misspelled in the `package.json` for this integration. I’m also a bit unsure about what’s going on because even the GitHub URL you suggest here doesn’t seem to match with the package published on npm 🤔 

In any case, the correct fix would be for the package author to publish a version with the correct URL.

PR #1199, opened 2024-07-31
Refactor `<ThemeGrid>` to automatically show themes added in a specific month

delucis:
This PR updates our `<ThemeGrid>` component used in monthly blog posts to automatially display themes for the specific month instead of requiring an array of all theme slugs.

Example in the latest blog post: https://deploy-preview-1199--astro-www-2.netlify.app/blog/whats-new-july-2024/

Usage:

```mdx
{/* Display themes published in July 2024 */}
<ThemeGrid date="2024-07-01" />
```

This is faster to author, but should also be faster to build as we now retrieve the theme metadata with a single API call instead of making a separate API call for every theme.

I updated past blog posts to use the new style of `<ThemeGrid>` retrospectively.

## Browser Test Checklist

I have only tested this PR on the following browsers so far:

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

(But this PR includes no DOM changes so probably OK, will check a few more soon)


sarah11918:
I compared the previews for the months here with the actual monthly blog posts, and June and July look identical.

Our actual May post had three themes, and this now picks up almost 30 of them. :sweat_smile: 
Same for April, this now captures about double the themes.

I don't think it's an issue for past posts, since they'll long have been read. Given that the two recent months match, I think this is fine going forward!

delucis:
Yes, I realised this will likely pick up some we didn’t feature initially, but I think that’s OK 😁

PR #1200, opened 2024-07-31
remove duplicate showcase entry

sarah11918:
Someone had added their website twice, once as a netlify.app and once under their own domain.

This removes the duplicate!


sarah11918:
Previewed both this month's blog post and the Showcase page and this seems fine to me!

delucis (changes requested):
We should probably also add the removed domain to the blocklist in our `update-showcase` script I think? Otherwise the site will be scraped again in the future.

LMK if you need me to take care of that!

sarah11918:
@delucis That would be helpful, yes! If you don't get to it today, I can do it later this evening.

delucis (approved):
Done. Should be good to go!

PR #1222, opened 2024-08-19
Fix startup on Gitpod

delucis:
This PR moves dependencies use for our showcase update script (notably `puppeteer` which has a heavy postinstall script) to `devDependencies` and everything else to `dependencies`.

Gitpod’s startup script as well as GitHub action configurations that don’t require them skip installing dev dependencies to speed them up/fix them in the case of Gitpod.

Also for other contributors, running `pnpm install --prod` will speed up dependency installs if it matters.

---

I initially tried using `optionalDependencies` for this but several things Astro depends on use those for platform-specific binaries, e.g. rollup and others list a dependency on a package for different OS and CPU architectures as optional and depend on npm to install just the one matching a user's platform. That means we can't skip installing optional dependencies because although named "optional", they are in fact being used as "just one of these please".

## Browser Test Checklist

n/a — repo configuration changes only


HiDeoo (approved):
Tested on Gitpod by appending `gitpod.io/#` to the PR URL, the installation works perfectly and everything seems to work as expected for me, great work :thumbsup:

PR #1225, opened 2024-08-21
Migrate authors collection to content layer

delucis:
Trying out the experimental content layer.

- Migrates the `authors` collection from a directory of small Markdown files to `src/data/authors/authors.json`.
  - `src/data/authors/` now includes all the avatar images too
  - Switched from using a custom `resolveImage` method to Astro’s `image()` utility, so the data file can now simply do `image: './username.png'` and have that resolved by Astro for us
  - A couple of bits of rendering logic needed updating to use `id` instead of `slug` and to remove use of the custom image resolution
  - ~~Hit a bug with the `file()` loader, which is currently patched in this PR but fixed properly upstream in https://github.com/withastro/astro/pull/11801~~ Updated Astro and no patches anymore 🎉 

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [x] Safari
- [ ] iOS Safari


src/content/config.ts
```diff
@@ -55,10 +59,8 @@ export const collections = {
 					'A date string or YAML date that is compatible with JavaScript’s `new Date()` constructor.',
 				),
 			authors: z
-				.array(
-					// biome-ignore format:
-					z.enum(['astro-team', 'ben', 'bjorn', 'chris', 'dan', 'drew', 'elian', 'ema', 'erika', 'fred', 'fuzzy', 'hideoo', 'jon', 'mark', 'martrapp', 'matt', 'matthew', 'nate', 'reuben', 'sarah', 'thuy', 'tony', 'yan']),
-				)
+				.enum(Object.keys(authors) as [keyof typeof authors, ...(keyof typeof authors)[]])
+				.array()
```
delucis, line 63: Switching authors to a single JSON file makes it easy to use an `enum` here which is locked to the file keys rather than maintaining this list twice.

tsconfig.json
```diff
@@ -6,7 +6,8 @@
 			"~/*": ["./src/*"]
 		},
 		"jsx": "preserve",
-		"jsxImportSource": "solid-js"
+		"jsxImportSource": "solid-js",
+		"resolveJsonModule": true
```
delucis, line 10: Enabled this to get proper types for the imported `authors.json` in `src/content/config.ts`

src/data/authors/authors.json
delucis, line 1: Opted for the somewhat clumsy `authors/authors.json` so we could colocate the data file with images more easily.

tsconfig.json
Princesseuh: our base config already includes this: https://github.com/withastro/astro/blob/735767f8dfe302834e48c2383c387b0e41c2f006/packages/astro/tsconfigs/base.json#L11-L12

delucis: Hmm, I could swear I changed something, but maybe I actually fixed it by switching from an array to an object.

matthewp (approved):

PR #1227, opened 2024-08-21
Automate homepage link with frontmatter field in blog posts

delucis:
This PR automates the homepage link to the latest blog post by adding a new `homepageLink` field to blog post frontmatter.

A blog post can include this and if it is the most recent post, it will automatically be used for the link on the landing page. For example, frontmatter can include something like

```yml
homepageLink:
  title: Astro 4.14
  subtitle: New experimental Content Layer API
```

And the homepage will render it as 

element with rounded corners reading Astro 4.14: New experimental Content Layer API, followed by an arrow pointing right

From what I could tell, we forgot to update this every single time we blogged for the last 2 months, so seems worth integrating it into the blog post frontmatter instead.

If a blog post is not considered important enough to link from the homepage, you can simply omit `homepageLink` and the post will be ignored.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [x] Safari
- [ ] iOS Safari

### DIFF
--- a/src/content/config.ts
+++ b/src/content/config.ts
@@ -49,6 +49,19 @@ export const collections = {
 				.describe(
 					'Summary of this blog post. Appears on the blog index as well as in metadata displayed on social media.',
 				),
+			homepageLink: z
+				.object({
+					title: z.string().max(16).describe('Very short call-out, e.g. `Astro 4.14` or `New!`.'),
+					subtitle: z
+						.string()
+						.max(50)
+						.optional()
+						.describe(
+							'Short tagline attracting attention to the post, e.g. `New experimental Content Layer API` or `Announcing Astro DB`. Avoid punctuation and keep things punchy.',
+						),
+				})
+				.optional()
+				.describe('Configure the homepage banner link for this post if it’s the most recent post.'),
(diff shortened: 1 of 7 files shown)


src/content/config.ts
```diff
@@ -49,6 +49,19 @@ export const collections = {
 				.describe(
 					'Summary of this blog post. Appears on the blog index as well as in metadata displayed on social media.',
 				),
+			homepageLink: z
+				.object({
+					title: z.string().max(16).describe('Very short call-out, e.g. `Astro 4.14` or `New!`.'),
```
sarah11918, line 54: Just checking that whatever we'll have planned re: `Astro 5.0 beta` will fit in 16 characters or less! :smile:

delucis: That’s 14 characters so… yes? 😄

sarah11918 (approved):
Preview is working for me on Chrome (Linux and Android)

src/content/config.ts
sarah11918: I counted it out lol! (The point is, I checked!)  :stuck_out_tongue:

PR #1244, opened 2024-09-04
Fix 404 link in Astro 4.14 blog post

crcastle:
Looks like the RFC number changed from 47 to 50.


delucis (approved):
Ah, great catch — thanks for the fix! 🚀

PR #1248, opened 2024-09-10
Modernize showcase collection

delucis:
This PR migrates our showcase collection to use more recent Astro features to simplify maintenance.

- Entries are now YAML data files instead of Markdown pretending to be data files
- Images are now resolved using the `image()` helper instead of a custom resolve function
- Images are now optimized with `astro:assets` — previously we committed two different resolutions to the repo and chose dynamically between them, now we just toggle the required width in `<Image>`
- Images are now directly colocated alongside the data files. Previously we have a `src/content/showcase/_images/` directory for these. Now each site’s image file should appear alongside its data file in the file browser, e.g.
   VS Code file tree showing the showcase directory and matching filenames in both WebP and YAML formats
   There still lots and lots of files — not so much I can do about that. (I decided against a single large `showcase.json` data file thinking it would make finding a specific entry or seeing new entries in Git a little more clumsy.)
- I’ve updated our `update:showcase` package script to match this new setup
- In the process I also replaced the `json-to-pretty-yaml` dependency with the battle-tested `yaml` one. `yaml` is already used by other dependencies, so although `json-to-pretty-yaml` is laudibly minimal and lightweight, its more limited features weren’t worth working around.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


sarah11918 (approved):
I browsed both a few pages from the showcase, and a blog post using the showcase entries and all looks good!

PR #1256, opened 2024-09-16
Remove `<BlogSummary>` component

delucis:
The `<BlogSummary>` component was added in #1003 to highlight the opening section. However, since then it has been incorrectly used on 15 occasions, resulting in odd markup and not the intended visual style.

This PR does the basic thing and rips it out entirely in favour of simply bolding the opening paragraphs like we already do in some other blog posts.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/components/BlogSummary.astro
+++ b/src/components/BlogSummary.astro
@@ -1,7 +1,0 @@
----
-
----
-
-<p class="text-xl">
-	<strong><slot /></strong>
-</p>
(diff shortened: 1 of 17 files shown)


TheOtterlord (approved):
Tested in Firefox

PR #1278, opened 2024-10-03
fix URL in Actions FeatureCard

martinklepsch:
Just noticed this URL isn't working anymore


delucis (approved):
Thanks for the fix @martinklepsch!

PR #1281, opened 2024-10-04
Add new "Utilities" category

Fryuni:
- Add a new "Utilities" category to the integration catalog for integrations with the keywords:
	- `tooling`
	-  `utils`
	-  `utility`
- Add a new "Loaders" category to the integration catalog for integrations with the keyword `astro-loader`
- Change the category fallback logic to add the catch-all `other` only for packages that didn't match any other category keyword. Previously, all packages would have the `css+ui` category because the `withastro` keyword required to join the catalog would not match any category.
- Updates the integration catalog by applying the changes above.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [x] Safari
- [ ] iOS Safari


delucis (changes requested):
Thanks @Fryuni!

I don’t seem to be able to review inline because GitHub is seizing up due to the many changed files (thank you GitHub), so leaving this as a single comment, sorry.

- Re: the labelling for the `other` category — I think “Uncategorized” would work best as the label here. Communicates a bit more clearly that this is the “junk drawer”

- Can you also open a matching docs PR to add the new keywords to https://docs.astro.build/en/reference/publish-to-npm/#collections ?

- Looks like you might need to re-run and update branch to resolve conflicts maybe?

Otherwise In general this looks good I think!

Fryuni:
> Re: the labelling for the `other` category — I think “Uncategorized” would work best as the label here. Communicates a bit more clearly that this is the “junk drawer”

Done

> Can you also open a matching docs PR to add the new keywords to [docs.astro.build/en/reference/publish-to-npm#collections](https://docs.astro.build/en/reference/publish-to-npm/#collections) ?

It was there from the beginning:
- https://github.com/withastro/docs/pull/9571

delucis (approved):
> It was there from the beginning

Of course it was 💜

Thanks again @Fryuni!

PR #1282, opened 2024-10-04
Fetch integration data in parallel for update

Fryuni:
- Update integrations in parallel, reduce the time for `update:integrations --unsafe` from 4 minutes to 10 seconds
- Use `p-limit` to avoid rate-limit errors by limiting the number of concurrent calls to NPM API to 10
	- `p-limit` is a already a dependency of Astro, so it is not an extra dependency to install


delucis (approved):
LGTM as long as we’re not getting rate limited! (I’m too used to GitHub’s aggressive, “do everything one-by-one or less” policy.)

PR #1285, opened 2024-10-07
Include `astro-integration` keyword in integrations catalogue

delucis:
We require the `astro-integration` keyword to make an npm package work with `astro add` ([docs](https://docs.astro.build/en/reference/integrations-reference/#allow-installation-with-astro-add)), but never actually tracked that in the integrations catalogue, requiring one of `withastro` or `astro-component` instead. This PR updates the integrations script to also treat `astro-integration` as a sufficient keyword to include a package in our catalogue and runs the update script to add a bunch of missing integrations.

Most notably this includes Expressive Code!

### DIFF
--- a/scripts/update-integrations.mjs
+++ b/scripts/update-integrations.mjs
@@ -105,7 +105,7 @@ async function fetchWithOverrides(pkg, includeDownloads = true) {
 }
 
 async function unsafeUpdateAllIntegrations() {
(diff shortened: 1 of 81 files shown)


HiDeoo (approved):
Great change imo :clap: (and also avoids me a PR to list one of my integration so even better 😄)

PR #1294, opened 2024-10-23
Update sponsor logos

clearlyTHUYDOAN:
## Description

Updates sponsor logos to reflect current status of official sponsors as well as gold sponsors

- Removes Storyblok as exclusive CMS sponsor
- Adds Manifest as new gold sponsor


clearlyTHUYDOAN:
Deploy shows that sponsor grid CSS might be more manual and we cannot simply remove the Storyblok logo. It currently looks like this after removing Storyblok and I'm afraid of ruining the CSS here.

My theory from inspecting the element is I have to update something related to the `grid-template-columns` or `grid-cols-x` related CSS or classes but I'm not sure @delucis. Any advice?

Screenshot 2024-10-22 at 10 29 29 PM

(As for the Manifest logo, I am currently asking them to re-supply an SVG with a transparent background.)

delucis (approved):
Thanks @clearlyTHUYDOAN!

I updated the classes on the sponsor grid to make it more tolerant of varying numbers of main sponsors. And I also found https://manifest.build/assets which included more SVG logos, so switched it out for one that looked more plausible, although you can change that again if they come back to you with something different.

Now looks like this:

image

PR #1305, opened 2024-10-31
Add `rel="nofollow ugc"` to links to external showcase sites and integrations

delucis:
Our showcase and integration catalogues link to a bunch of external resources.

This PR adds `rel="nofollow ugc"` to these to reduce their value to the external site in terms of SEO (see [“Qualify your outbound links to Google”](https://developers.google.com/search/docs/crawling-indexing/qualify-outbound-links)).

We’ve had a few signs of potential inauthentic activity using these catalogues, so worth adding these I think to discourage that.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/components/charts/CWVCLSChart.astro
+++ b/src/components/charts/CWVCLSChart.astro
@@ -5,7 +5,7 @@ import CardLabel from '../CardLabel.astro';
 ---
 
 <Card className="mx-auto max-w-2xl">
-	<CardBody>
+	<CardBody :any>
 		<CardLabel>% Sites Passing CLS</CardLabel>
 		<canvas id="chart-cwv-cls"></canvas>
 	</CardBody>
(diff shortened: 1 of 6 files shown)


delucis:
> Should we add `ugc` too?

Yeah, I considered it. Not sure if it makes a difference or not?

src/components/charts/CWVCLSChart.astro
```diff
@@ -5,7 +5,7 @@ import CardLabel from '../CardLabel.astro';
 ---
 
 <Card className="mx-auto max-w-2xl">
-	<CardBody :any>
```
Princesseuh, line 8: What was that?

delucis: Turns out it’s needed?? See typecheck

ascorbic (approved):

PR #1315, opened 2024-11-02
Integrations update

### LINKED ISSUE #1313: Integration: some homepage urls are wrong
https://astro.build/integrations/ lists all the npm package related to astro.
When selecting an integration, this links to the url page describing the integration.
Most of the time (apart when an issue has been filled to overwrite it), this url is the one found in package.json as "homepage".

Unfortunately, many of these urls are wrong. 48 "404" errors on homepage url, on a total of 690 integrations, which leads to a poor user experience.

As an example: integration "astro-github-stats" has the homepage https://ayco.io/showcase/astro-github-stats, which cannot be accessed. I guess it used to work at some point, but this is no more the case.

I propose: 
* to update scripts/update-integrations.mjs with an option to list wrong homepages
* update wrong urls with the npm url of the package. In the case of astro-github-stats, https://www.npmjs.com/package/astro-github-stats provides relevant information. This update could be done by the script, but a manual check may be useful. That's why I'm not sure whether such a script should be run on nightly or weekly build.

Let me know what you think of this, and if I should go toward this direction.

Do not hesitate to close this issue if you think this is dumb or if this was already discussed.

pascal-brand38:
## Browser Test Checklist

- [X] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

Serie of 4 commits to fix close #1313.
It consists of an update of scripts/update-integrations.mjs:
- use of yargs to manage command-line options of the script
- optimization of safeUpdateExistingIntegrations()
- adding the option fix-homepageurl to update wrong homepage url of integrations
- fix src/content/integration/**/*.md files using the result of the script **npm run update:integrations -- --fix-homepageurl**

Because #1313 is still in discussion, and as this script is used in nightly and weekly builds, I think special attention has to be paid. This is why this pull-request is open as a draft.

Thanks for your help!


scripts/update-integrations.mjs
```diff
@@ -15,6 +15,38 @@ import {
 } from './integrations.mjs';
 import { markdownToPlainText } from './markdown.mjs';
 import { fetchDetailsForPackage, fetchDownloadsForPackage, searchByKeyword } from './npm.mjs';
+import _yargs from 'yargs'
+import { hideBin } from 'yargs/helpers';
```
delucis, line 19: I think `yargs` is probably overkill here — this script is run by CI tasks 99.9% of the time. I think instead of adding an extra `fix-homepageurl` option it should just be part of the `unsafe` command. There aren’t really scenarios where you _don’t_ want that right? That way it’s still just a single flag and we do without `yargs`.

scripts/update-integrations.mjs
```diff
@@ -199,14 +231,55 @@ Updated: ${existingIntegrations.size - deprecatedIntegrations.length} integratio
 	console.info(stats);
 }
 
-async function safeUpdateExistingIntegrations() {
+async function urlStatus (url) {
+	try {
+		await axios.head(url, {timeout: 10000});
```
delucis, line 236: Similarly, not sure we need to add `axios` as a dependency here. We can achieve similar with a native `fetch(url, { method: 'HEAD' })`. Doesn’t throw an error or timeout, but can still check the status and could even be made to timeout with an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) if that’s very important (speed isn’t essential to these scripts though IMO, so definitely OK to skip that if you prefer).

delucis (commented):
Thanks @pascal-brand38! Left a couple of preliminary notes regarding the added dependencies. (These are also breaking the Netlify build due to the lock file being out of date, but I think we may not need them anyway.)

delucis (commented):
Left one small note, but looks good to me — thanks @pascal-brand38!

scripts/update-integrations.mjs
```diff
@@ -139,6 +139,19 @@ async function unsafeUpdateAllIntegrations() {
...
+				// if not, replace it by the link to the package on npm
+				try {
+					const response = await fetch(details.homepageUrl, { method: 'HEAD'})
+					if (response.status === 404) {
+						details.homepageUrl = `https://www.npmjs.com/package/${data.name}`
+					}
+				} catch (error) {
+					// such an error may occur when the hostname is unknown
+					// we consider it ok at this may come from a down server
+					console.info(`Package ${details.name}: Cannot access ${details.homepageUrl} (not fixed)`)
```
Fryuni, line 152: I'd expect this to happen more often than 404s, when a website is shut down or a domain no longer resolves. 

For the user experience, it is equally detrimental to see a 404 page or a browser error page. So I think this case should also point to NPM.

If the service is just down temporarily it will be updated back in the following weekly run of the script.

delucis (approved):
Thanks again for taking care to add this @pascal-brand38! I’ll merge this and run the “unsafe” script now to test it out 🚀

PR #1347, opened 2024-12-03
Upgrade Astro to v5

delucis:
This PR updates dependencies for the astro.build repo, including updating for Astro v5.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


HiDeoo:
Only for information, the type issue seems to be related to https://github.com/withastro/astro/issues/12598.

> `getEntry` no longer returns an entry or `undefined` but an entry, so as the code is `(await getEntry('blog', slug)) || (await getEntry('caseStudies', slug))`, TypeScript sees that the first `getEntry` will always return an entry so `post.collection` is always `blog`.

sarah11918 (approved):
Confirming that I previewed in Chrome, browsed the blog and showcase, filtered some themes and integrations, and checked on the wallpapers and nothing looks out of place that I noticed!

PR #1349, opened 2024-12-12
replace studio with sponsors

sarah11918:
replaces the footer link to Studio with a link for "Sponsors" that right now goes to the Open Collective page.

In future, we may want a dedicated Sponsors page so that we can offer placement here as a benefit for certain sponsorship amounts. For now, the open collective is both a call to action to sponsor and our list of sponsors.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/components/Footer.astro
+++ b/src/components/Footer.astro
@@ -36,12 +36,12 @@ const groups: LinkGroup[] = [
 	{
 		title: 'Ecosystem',
 		links: [
-			{ label: 'Studio', href: 'https://studio.astro.build/' },
 			{ label: 'Community', href: 'https://community.astro.build' },
 			{
 				label: 'Contributing',
 				href: 'https://github.com/withastro/astro/blob/main/CONTRIBUTING.md',
 			},
+			{ label: 'Sponsors', href: 'https://opencollective.com/astrodotbuild#category-CONTRIBUTE' },
 		],
 	},
 	{


src/components/Footer.astro
```diff
@@ -36,12 +36,12 @@ const groups: LinkGroup[] = [
 	{
 		title: 'Ecosystem',
 		links: [
-			{ label: 'Studio', href: 'https://studio.astro.build/' },
 			{ label: 'Community', href: 'https://community.astro.build' },
 			{
 				label: 'Contributing',
 				href: 'https://github.com/withastro/astro/blob/main/CONTRIBUTING.md',
 			},
+			{ label: 'Sponsors', href: 'https://opencollective.com/astrodotbuild#category-CONTRIBUTE' },
```
delucis, line 44: Should this not include the link to a sub section?
```suggestion
			{ label: 'Sponsors', href: 'https://opencollective.com/astrodotbuild' },
```

delucis (approved):
Thanks for noticing this! Left one query but the idea makes sense to me!

src/components/Footer.astro
sarah11918: My thinking was that taking it to contribute is the call to action, plus that's where our sponsors themselves are most visible, but comparing that to the main link it basically gets you right there anyway. Sound good!

PR #1228, opened 2024-08-22
Accessibility and performance fixes

delucis:
This is a bit of an omnibus PR tackling many small paper cuts around accessibility and performance on astro.build. That makes it a little tricky to review, but here are the changes broken down by where they were made. For the most part nothing has changed visually, or if it has, in very tiny ways I would expect no-one to notice — I’ve marked these with 👁️.

#### Miscellaneous

- Adds missing accessible labels to UI and images in the gallery components for individual theme pages, e.g. [/themes/details/portfolio/](https://deploy-preview-1228--astro-www-2.netlify.app/themes/details/portfolio/)
- Fixes heading levels in some pages that were jumping `h1` => `h3`, e.g. [/privacy/](https://deploy-preview-1228--astro-www-2.netlify.app/privacy/)
- Fixes some card grid usage where the structure was `ul > article`. These are now `ul > li > article`, e.g. [/press/](https://deploy-preview-1228--astro-www-2.netlify.app/press/#assets)
- Uses trailing slashes on some internal theme and showcase links to match canonical URLs (showcase pages were redirecting, so this change improves performance; theme pages were not so this improves SEO)

#### `/db/` landing page

- Fixes [use-case tabs on `/db/`](https://deploy-preview-1228--astro-www-2.netlify.app/db/#use-cases) to address incorrect `aria-*` attributes (also reduces JS quite a bit because the custom element was managing a bunch of unnecessary attribute state).
- Fixes text contrast in [`/db/` features section](https://deploy-preview-1228--astro-www-2.netlify.app/db/#features). (👁️ This is a visual change that does make the illustrations impacted slightly less subtle, but text contrast requirements still need to be met.)
- Fixes accessible labelling of secondary navbar in `/db/`.
- Fixes miscellaneous misused elements and attributes in `/db/` (e.g. `<dl>` used without `<dt>` and `<dd>`)

#### Homepage

- Uses `div > section[role="tabpanel"]` for the tabbed interfaces on the landing page instead of `ul > li[role="tabpanel"]`. The `tabpanel` role removes the list-item semantics from an `li` making it invalid to use inside a `ul`.
- Reduces the amount of blur used on background image in the hero section of the landing page. Safari in particular was really struggling to render the 64px blur over such a large area. Reducing the blur amount has minimal visual impact while fixing performance.
- Sets `loading="lazy"` on images below the fold on the landing page to improve page load performance.
- Uses an `` instead of a background image for the background of the homepage hero element to prioritise its loading and improve perceived page performance.
[...]

PR #1350, opened 2024-12-13
Remove large blurs

### LINKED ISSUE #1274: astro.build main page performance is awful in Safari
There are severe performance issues in Safari on the main page. Most likely because of expensive relayout involving too many SVG elements on the page. Hover animations, which are causing relayout, are taking up to 1 second.

Safari 17.5
MacBook Pro 2019 2.6Hz 6-Core Intel Core i7

delucis:
This PR aims to close #1274 

In several places we had a very large (160px) blur on decorative background elements. These look great, but do not perform well across browsers. In #1228 I had already reduced blur on the hero image to just 16px to improve Safari performance, but didn’t notice these other large blurs at the time.

This PR

- Replaces the two instances of 160px blur with an alternative approach using a radial-gradient mask on the element to simulate a blurred fall-off of intensity. To the trained eye, these are subtly different effects, but they are close enough and I think the trade-off is justified to address poor performance.
- Removes several places we were using `backdrop-blur` unnecessarily (the backdrop was already something blurry or a flat colour so recalculating additional blurs was superfluous)
- Removes an old component we were no longer using

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [x] Safari
- [ ] iOS Safari


sarah11918 (approved):
I am only on Chrome and can't perceive a difference between the deploy preview and live docs, so certainly nothing appears to have gone wrong in this PR for me!

PR #1365, opened 2025-01-06
Ensure showcase titles are not "Just a moment..."

HiDeoo:
As discussed on Discord, this PR ensures that the showcase titles are not "Just a moment..." by enforcing that in the `showcase` collection schema.

This title usually means that fetching infos was blocked, e.g. on Cloudflare hosted sites and that the title will need to be updated manually.

Happy to discuss the wording of the error message as usual.

## Test Checklist

I have tested this PR by building the website locally to ensure no entries have the title "Just a moment...".

I also changed the title of one of the showcase entries to "Just a moment..." to ensure that the build fails.

image

## Browser Test Checklist

- [ ] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

This change does not affect the browser output.

### DIFF
--- a/src/content/config.ts
+++ b/src/content/config.ts
@@ -162,7 +162,13 @@ export const collections = {
 		type: 'data',
 		schema: ({ image }) =>
 			z.object({
-				title: z.string().min(1),
+				title: z
+					.string()
+					.min(1)
+					.refine((value) => value !== 'Just a moment...', {
+						message:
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks for jumping on this! And I think the error message looks perfect — clearly indicates the issue to me.

PR #1366, opened 2025-01-06
Implement author profiles

TheOtterlord:
<details>
<summary>screenshots</summary>

image
image

</details>

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [ ] Safari
- [ ] iOS Safari


src/pages/themes/author/[username]/[...page].astro
```diff
@@ -0,0 +1,78 @@
...
+const currentPage =
+	typeof Astro.params.page === 'undefined' ? 1 : Number.parseInt(Astro.params.page);
+
+// invalid page number!
+if (!currentPage || Number.isNaN(currentPage)) {
+	return Astro.redirect('/404');
+}
+
+const allThemes: ThemeAndAuthor[] = await fetch(
+	`${THEMES_API_URL}/api/themes/author?username=${Astro.params.username}`,
```
Comment (delucis) [inline src/pages/themes/author/[username]/[...page].astro:28]: Just noting how we discussed maybe not to use `username` as the param here so we don’t expose GitHub accounts that maybe were not intended to be public like this.

One proposal was to use the numeric author ID instead. Would require updating the API endpoint too.

src/pages/themes/_components/ThemeCard.astro
delucis, line 1: I don’t think we should link the names on these theme cards — the entire card reacts on hover and appears to be clickable as a single entity, so I think introducing different click behaviours to different parts of the card could be confusing to people.

In addition, as currently implemented, it introduces “holes” in the clickable area where clicking has no effect.

While testing this PR, these links can feel useful, but I think in practice the link on individual theme pages is the key place where people are likely to think “this looks nice, wonder what else they made” more than the overview list which is all about picking themes rather than authors. If later we wanted to surface authors themselves more prominently as an entry point then that could be done via some alternative catalogue UI (e.g. “top authors” etc.)

---

We probably _should_ fix these cards to not wrap the full card content in a link too, but that can happen in a separate, a11y-focused PR (basically the ideal pattern here is: card title is the link so the accessible list of links is concise and descriptive, but then use `a::before` or `a::after` absolute positioned over the full card so it still behaves the same way when clicked, but I digress…)

src/components/Profile.astro
```diff
@@ -0,0 +1,56 @@
...
+			}
+			</small>
+		</div>
+		<hr class="border-astro-gray-500" />
+		<div class="flex items-center justify-between">
+			<small class="code text-astro-gray-200">GitHub</small>
+			<small class="text-astro-gray-200">
+				<a class="link" href={`https://github.com/${author.username}`}>@{author.username}</a>
+			</small>
+		</div>
```
delucis, line 41: For the same reasons as discussed re: the URLs, I think we should probably omit this:

```suggestion
```

src/components/Profile.astro
```diff
@@ -0,0 +1,56 @@
...
+const { author } = Astro.props;
+---
+
+<div class="flex lg:flex-col flex-wrap items-center gap-4">
+	<Image
+		inferSize
+		src={author.avatar ?? ''}
+		height={48}
+		width={48}
+		class="rounded-full size-12 lg:w-full h-fit border"
```
delucis, line 18: I think omitting the border and rounding can make this gel more with the general site feel:
```suggestion
		class="size-12 lg:w-full h-fit"
```
The astro.build design is very “square” in general, reliant on hard edges and borders except for smaller design details (like pills, buttons, and avatars bubbles).

Here are a few examples of what this looks like:

| Desktop | Mobile |
|---|---|
| Screen Shot 2025-01-09 at 23 38 20 | Screen Shot 2025-01-09 at 23 38 41 |
| Screen Shot 2025-01-09 at 23 37 53 | Screen Shot 2025-01-09 at 23 39 07 |
| Screen Shot 2025-01-09 at 23 37 26 | Screen Shot 2025-01-09 at 23 38 54 |

src/components/Profile.astro
```diff
@@ -0,0 +1,56 @@
...
+					<small class="code text-astro-gray-200">Website</small>
+					<small class="text-astro-gray-200">
+						<a class="link" href={author.url}>
+							{author.url}
+						</a>
+					</small>
+				</div>
+			)
+		}
+	</div>
```
delucis, line 55: I think the text size in this sidebar section may need tweaking. I’d use the right sidebar on theme pages as a reference.

For comparison:

| | |
|---|---|
| image | image |

In the theme page sidebar, values are slightly larger.

src/components/Profile.astro
```diff
@@ -0,0 +1,56 @@
...
+			</small>
+		</div>
+		{
+			author.url && (
+				<hr class="border-astro-gray-500" />
+				<div class="flex items-center justify-between">
+					<small class="code text-astro-gray-200">Website</small>
+					<small class="text-astro-gray-200">
+						<a class="link" href={author.url}>
+							{author.url}
```
delucis, line 49: I’d recommend stripping protocol and trailing slashes here for legibility, i.e. `astro.build` vs `https://astro.build/`.

delucis (commented):
Very cool work @TheOtterlord! Left some notes based on our discussions and also a couple of design thoughts.

src/pages/themes/author/[username]/[...page].astro
Comment (TheOtterlord) [reply src/pages/themes/author/[username]/[...page].astro]: Implemented in this PR. Making another PR to dev portal to implement the API change

src/components/Profile.astro
TheOtterlord: Removed the small tag, but the link styles will still make it slightly darker and give the on hover effect

src/components/Profile.astro
TheOtterlord: Looks great!

src/pages/themes/_components/ThemeCard.astro
TheOtterlord: Sounds good. Will spin this into an issue tomorrow.

delucis (commented):
Looking great @TheOtterlord! And cool to see those articles properly labelled in the accessibility tree:
Chrome dev tools view of a group of article nodes with distinct labels

Spotted a few tiny details but otherwise looking great!

src/components/Profile.astro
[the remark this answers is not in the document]
TheOtterlord: Moved!

delucis:
Thanks for the updates @TheOtterlord! Turns out I was wrong about the 404s I guess. It works locally but in the preview deploy I just get a blank page. Guess we should figure out how to do that properly. 

One other 404 thing: looks like we need to handle pagination 404s. Currently I get content even if I add an out of range page number e.g. https://deploy-preview-1366--astro-www-2.netlify.app/themes/author/823/

src/pages/themes/author/[id]/[...page].astro
```diff
@@ -0,0 +1,82 @@
...
+export const prerender = false;
+
+// with '[...page]' rest routes we'll get undefined for the first page, default that to 1
+// otherwise, try to parse the page number from the URL
+const currentPage =
+	typeof Astro.params.page === 'undefined' ? 1 : Number.parseInt(Astro.params.page);
+
+// invalid page number!
+if (!currentPage || Number.isNaN(currentPage)) {
+	return new Response(null, { status: 404 });
```
Comment (delucis) [inline src/pages/themes/author/[id]/[...page].astro:24]: OK, chatted a bit with some of the framework team and seems we don’t have an ideal solution here yet — the above _should_ work, but because our 404 page is static, it is not available to the server at runtime. That’s why it works in dev currently but not in production.

Given that I _think_ the best of our options is to return to using a redirect like you initially had:

```suggestion
	return Astro.redirect('/404/', 404);
```

At least means a user gets a nice 404 page and we serve the correct status code. The future enhancement would be to switch to a technique that doesn’t alter the URL once that’s supported.

delucis (commented):
OK, sorry for the churn but recommending we switch back to `Astro.redirect()` after all 😅 

After that, should be good to go I think!

src/pages/themes/author/[id]/[...page].astro
Comment (TheOtterlord) [reply src/pages/themes/author/[id]/[...page].astro]: Used in two places, so made a commit

PR #1383, opened 2025-01-27
Small design detail tweaks

delucis:
This PR fixes a few tiny design details in a few spots I’ve noticed while reviewing other work the last few days.

- Fixes some awkward wrapping of the bottom part of theme cards in the theme catalogue.
  
  | Before | After |
  |---|---|
  | three theme cards, their footer text wraps awkwardly in the narrower layout | the same theme cards, now their footer text sits on a single line, truncated with ellipses when necessary

- Fixes some slightly tight spacing around the “used by” company logos on the homepage. This change also ensures we never have a layout where just one logo is wrapped by itself on to the second row.

  | Before | After |
  |---|---|
  | row of logos running right to the edge of the browser window | logos wrapping with plenty of space around them |
  | same logos, the last one wraps to a new row by itself | same logos as before, wrapping in a balanced way |

- Improves layout for the large page title text on showcase, integration, and theme catalogue pages. Currently this often wraps just a single word onto a new line. This change applies `text-wrap: balance` to ensure a balanced composition when the text must wrap and also increases the minimum width so it doesn’t have to wrap on larger screens.

  #### Examples
  | Before | After |
  |---|---|
  | image | image |
  | image | image |
  | image | image |

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/pages/_components/landing-page/Teams.astro
+++ b/src/pages/_components/landing-page/Teams.astro
@@ -11,13 +11,13 @@ const logos = [
 ] satisfies Array<{ icon: string; label: string; classes?: string }>;
 ---
 
-<div class="mt-8 sm:mt-16 xs:p-6 mx-auto w-full max-w-screen-xl space-y-5 sm:space-y-8 p-4 sm:p-8">
+<div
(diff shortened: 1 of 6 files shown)


sarah11918 (approved):
I didn't always see all the "before" issues, but everything looks good in the preview to me! Chrome desktop

PR #1387, opened 2025-01-30
feat: use external redirects for socials

ematipico:
This PR enables external redirects for our socials, essentially removing the need for `/on/[platform].astro` page.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


src/pages/on/[platform].astro
```diff
@@ -1,20 +0,0 @@
...
-		params: { platform: link.platform },
-		props: { link },
-	}));
-}
-
-const mastodon = siteInfo.socialLinks.find((link) => link.platform === 'mastodon');
-const { link } = Astro.props;
----
-
-{mastodon && <link rel="me" href={mastodon.me ?? mastodon.href} />}
```
Comment (ascorbic) [inline src/pages/on/[platform].astro:19]: Is removing this going to break anything?

Comment (natemoo-re) [reply src/pages/on/[platform].astro:19]: Oh that’s a great catch. Technically this is how Mastodon verifies ownership of our domain. We could add the `rel=me` to the BaseHead component instead.

natemoo-re (approved):
Much better 🙌

pnpm-lock.yaml
```diff
@@ -1241,21 +1247,39 @@ packages:
...
 
+  '@shikijs/themes@1.29.2':
+    resolution: {integrity: sha512-i9TNZlsq4uoyqSbluIcZkmPL9Bfi3djVxRnofUHwvx/h6SRW3cwgBC5SML7vsDcWyukY0eCzVN980rqP6qNl9g==}
+
   '@shikijs/types@1.27.2':
     resolution: {integrity: sha512-DM9OWUyjmdYdnKDpaGB/GEn9XkToyK1tqxuqbmc5PV+5K8WjjwfygL3+cIvbkSw2v1ySwHDgqATq/+98pJ4Kyg==}
 
+  '@shikijs/types@1.29.2':
+    resolution: {integrity: sha512-VJjK0eIijTZf0QSTODEXCqinjBn0joAHQ+aPSBzrv4O2d/QSbsMw+ZeSRx03kV34Hy7NzUvV/7NqfYGRLrASmw==}
+
```
delucis, line 1282: Might be worth running `pnpm dedupe` to clean up some of these duplicates here. (Same for Vite which is getting included twice too.)

src/pages/on/[platform].astro
Comment (delucis) [reply src/pages/on/[platform].astro:19]: Ah, I’m not even sure that would be sufficient for what this is doing right now?

For background: Mastodon uses a page’s `<link rel="me">` to verify that that page belongs to a specific Mastodon account. You can see this on [our profile](https://m.webtoo.ls/@astro) currently:

Astro Mastodon profile

The links are all displayed in green as “verified” because Mastodon visits e.g. `astro.build/on/bluesky` and sees the `<link>` tag pointing back to our Mastodon profile.

Does that mean we actually have to stick with our current approach after all? Or adding the link to our BaseHead validate that we own all of `astro.build` including the `/on/...` links? (My guess would be visiting `/on/...` now redirects straight to Bluesky etc. so they don’t appear to be owned by us?)

delucis (commented):
Thanks @ematipico!

src/pages/on/[platform].astro
Comment (ascorbic) [reply src/pages/on/[platform].astro:19]: Could we add a static page just for mastodon?

Comment (ascorbic) [reply src/pages/on/[platform].astro:19]: oh I see what you're saying. So no, we can't

ematipico:
I suppose we can just apply the workaround for Mastodon until core allows to customise the redirect page

delucis:
I still don’t think this will work because we need the `<meta>` tag for all of these links, so will close this. Sorry @ematipico that I didn’t realise we had this setup earlier!

PR #1401, opened 2025-02-15
Add Starlight 0.32 blog post

delucis:
Adds a blog post for Starlight’s v0.32 release

## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


src/content/blog/starlight-032.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+- [**Route data middleware**](#a-new-route-data-paradigm)
+- [**New i18n APIs for plugins**](#new-i18n-apis-for-plugins)
+- [**Updated plugin hooks**](#updated-plugin-hooks)
+- [**Multisite search support**](#multisite-search-support)
+
+To upgrade an existing Starlight site, use the automated `@astrojs/upgrade` CLI tool. This will update Starlight, Astro, and any other integrations you are using:
+
+```sh
+npx @astrojs/upgrade
```
HiDeoo, line 23: Considering this may break many plugins, should we add a mention that plugins are not handled by `@astrojs/upgrade`?

src/content/blog/starlight-032.mdx
```diff
@@ -0,0 +1,154 @@
...
+```
+
+## A new route data paradigm
+
+Historically, users and plugin authors have found themselves reaching for Starlight’s [component override](https://starlight.astro.build/guides/overriding-components/) system when all they really wanted to do was tweak the data rendered by Starlight’s default components.
+
+In v0.32, we are introducing **a new route middleware system** that gives users full access to Starlight’s data model without needing to override any components. This is particularly helpful for plugins as it avoids common conflicts where several plugins tried to override the same component.
+
+Similar to Astro’s middleware system, route middleware is called for every Starlight page render and gives you the opportunity to modify any data before it gets rendered. This is a powerful way to implement custom logic not currently possible with configuration alone.
```
HiDeoo, line 32:
```suggestion
Similar to Astro’s middleware system, a route middleware is called for every Starlight page render and gives you the opportunity to modify any data before it gets rendered. This is a powerful way to implement custom logic not currently possible with configuration alone.
```

Is this missing "a" or maybe "route middlewares are"? Or maybe it's fine like this, just sounded weird to me.

src/content/blog/starlight-032.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+To upgrade an existing Starlight site, use the automated `@astrojs/upgrade` CLI tool. This will update Starlight, Astro, and any other integrations you are using:
+
+```sh
+npx @astrojs/upgrade
+```
+
+## A new route data paradigm
+
+Historically, users and plugin authors have found themselves reaching for Starlight’s [component override](https://starlight.astro.build/guides/overriding-components/) system when all they really wanted to do was tweak the data rendered by Starlight’s default components.
```
HiDeoo, line 28: We mention that a route middleware is a better place for some operations, but should we say, maybe at the end, why an user may want to still reach to an override?

src/content/blog/starlight-032.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+This example plugin logs Starlight’s default “Built with Starlight” text string and uses the translation matching the user’s locale if it is available:
+
+```js "useTranslations" "useTranslations(userLocale)"
+export default {
+	name: 'localizedPlugin',
+	hooks: {
+		'config:setup'({ useTranslations, logger }) {
+			// Detect the current user’s preferred locale.
+			const userLocale = Intl.DateTimeFormat().resolvedOptions().locale;
```
HiDeoo, line 71: This is really neat 👍

HiDeoo (commented):
Writing is not my strong suit but I personally like the general flow of the article, only left a few minor comments :clap:

src/content/blog/starlight-032.mdx
delucis: Great idea — done in c08a69490!

src/content/blog/starlight-032.mdx
delucis: Very good question 😅

I… actually have no idea now you ask this. For some reason in my head “middleware” sounds like a plural/collective noun, e.g. often people will say things like “this website uses middleware” rather than “a middleware” or “middlewares”. So here it is describing the full stack of middleware being called not a specific one. 

Looking around I think this aligns with its related words like “software” which also doesn’t usually get pluralized to “softwares”. The [Wiktionary definition](https://en.wiktionary.org/wiki/middleware) also describes it as “usually uncountable”, which I guess means the same thing?

src/content/blog/starlight-032.mdx
delucis: Made an attempt in 130e765fd to reframe this a bit so that the distinction between the two use cases is hopefully a bit clearer.

HiDeoo: Really like the new version personally, I think it's way more obvious why you would want to pick one or the other approach 👍

HiDeoo (approved):
Personally very happy with the post, great work :clap:

sarah11918 (approved):
Looks great! :partying_face:

PR #1413, opened 2025-02-25
Refactor icon usage to consistently use `<Icon>`

delucis:
Currently, we render icons with a mixed bag of `.tsx` Solid components and SVGs used with the `<Icon>` component from `astro-icon`. The Solid components were once upon a time created with the intention that they could be used in interactive client-side components, but we don’t appear to be doing that, so this PR consolidates all usage around the `<Icon>` component and removes Solid entirely (it’s not used anywhere except for icons).

This is a preliminary clean-up PR but I was mainly motivated to clean up before upgrading to use the v1 of `astro-icon` — we’re still on an old v0 release.

There should be zero visual impact from this PR (small white lie — there is one icon usage that shifted by 2px in an unimportant way, but I dare you to spot where).

As part of removing Solid, I also switched over our use of Solid’s signals (the only runtime Solid code we were using) for Preact’s standalone `@preact/signals` package. These are functionally equivalent just with a slightly different API, but the Preact version has a couple of benefits:

- It is a standalone package, so its install size is ~95% smaller than Solid’s (184 KB vs 3.4 MB).

   `solid-js`: <a href="https://pkg-size.dev/solid-js"></a>
   
   `@preact/signals-core`: <a href="https://pkg-size.dev/@preact/signals-core"></a>

- Bundled, the client-side runtime for Preact ~50% smaller than the Solid equivalent (1.4 kB vs 2.7 kB gzipped)

   dev tools network panel showing 2.7 kB compressed size for Solid

   dev tools network panel showing 1.4 kB compressed size for Preact signals

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


delucis:
Whelp, I’d missed we were using Solid’s signals implementation somewhere. Will need to fix that tomorrow.

delucis:
> Whelp, I’d missed we were using Solid’s signals implementation somewhere. Will need to fix that tomorrow.

Fixed by switching to Preact signals rather than reinstalling Solid — see main PR comment for details.

sarah11918 (approved):
Checked every page on ChromeOS laptop and don't see anything wrong!

PR #1380, opened 2025-01-23
Information architecture revamp

delucis:
This PR reworks our navigation structure to help better surface content.

- Adds a “Community” menu to our global navigation to better surface our sponsor and merch pages as well as clearly call out Discord as a community location.
- Adds a new top-level case studies landing page and links this in our “Resources” menu so that it is easier to find case studies (currently these are buried in the blog feed and it’s not obvious they exist).
- Reorganizes the footer nav for a slightly cleaner grouping and increases the social icons sizes so they are a bit more visible.
- There are a few small bits of refactoring/fixing stuff around here too. For instance, I noticed our dropdown menus were using a cooler black background than we generally do, which is subtle, but impossible to unsee 😅 
   dropdown menu a subtly different shade from the surrounding page background
   So that’s fixed here.

## Browser Test Checklist

I have tested this PR on at least three of the following browsers:

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

PR #1414, opened 2025-02-27
Update Astro Icon to v1

delucis:
Updates `astro-icon` to v1 following their [upgrade guide](https://www.astroicon.dev/guides/upgrade/v1/)

Other changes they don’t document in the upgrade guide I encountered:

- v1 no longer prefixes IDs in SVGs by default, which can lead to clashes that break things (e.g. if two icons use a gradient with `id="a"`, these will conflict), so I had to manually add a random prefix to IDs using the SVGO configuration options.

- v1 seems to always add a `width` attribute, which wasn’t previously the case. This broke some of our responsive icon sizing in a couple of places where I needed to use and explicit `width="auto"` attribute to combat the change.

Would appreciate more eyes going through the previews on this one in case there’s a spot I missed where an icon is displayed incorrectly.

Main benefit of the upgrade apart from just not depending on the older release is the improvements in how Astro Icon reuses duplicate icons. It’s not a massive difference but still, on the homepage it’s ~2 kB gzipped less over the wire for example.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


sarah11918:
The only thing I spotted was the `/press/` page looking wonky: 
image

src/components/Footer.astro
```diff
@@ -1,5 +1,5 @@
 ---
-import Icon from 'astro-icon';
+import { Icon } from 'astro-icon/components';
```
HiDeoo, line 2: Unrelated, but looks like none of the `Link` in `LinkGroup` use an icon anymore.

src/pages/press/index.astro
HiDeoo, line 1: I think there are a few issues with the icons for the old/new logo in the following sections:

- Logo history
- Spacing
- Minimum size

HiDeoo (commented):
Checked locally all changed components and MDX content, and also against the upgrade guide. Only spotted 1 issue 👍

src/components/Footer.astro
delucis: Ah yeah, I briefly had social icons in there as part of #1380 but then removed that, but looks like I forgot to also remove the type change — thanks!

src/pages/press/index.astro
delucis: Thank you for spotting! Fixed in 36270d3.

astro.config.mjs
```diff
@@ -23,6 +24,18 @@ export default defineConfig({
 			themes: [houston],
 			frames: false,
 		}),
+		icon({
+			svgoOptions: {
+				plugins: [
+					{ name: 'preset-default' },
+					{
+						name: 'prefixIds',
+						// Ensure IDs used in SVGs are unique to avoid clashes between inline SVGs.
```
natemoo-re, line 33: Ugh, good catch. Will fix this upstream!

natemoo-re (approved):
LGTM! Also tested on iOS Safari and didn't see anything broken.

HiDeoo (approved):

astro.config.mjs
delucis: There is this open issue I found when I hit this: https://github.com/natemoo-re/astro-icon/issues/127

PR #1469, opened 2025-03-21
Ensure that themes with exactly 2 images display both images

HiDeoo:
This PR fixes an issue where themes with only 1 extra image would not display the extra image.

The default image is defined in the `image` field of a `Theme` object and extra images are defined in the `images` array. The current condition to display the image gallery component is `theme.Theme.images.length > 1` which is incorrect as it means that the gallery component will not be displayed if there is only 1 extra image.

This PR fixes the issue by changing the condition to `theme.Theme.images.length > 0` so that the gallery component will be displayed if there is at least 1 extra image (and bring it on par with the condition [used in the portal UI](https://github.com/withastro/astro-dev-portal/blob/c375f9e2d7b6118cfafc78587dff503e0ea4190e/apps/site/src/pages/themes/%5Bslug%5D.astro#L85))

## Preview

A preview of the fix can be seen with the Starlight Flexoki theme that has only 1 extra image.

- Current version: https://astro.build/themes/details/starlight-flexoki
- Preview with the fix: https://deploy-preview-1469--astro-www-2.netlify.app/themes/details/starlight-flexoki/

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [x] Safari
- [x] iOS Safari

### DIFF
--- a/src/pages/themes/details/[slug].astro
+++ b/src/pages/themes/details/[slug].astro
@@ -32,7 +32,7 @@ const relatedThemes: Array<ThemeAndAuthor> = await fetch(
 >
 	<div class="grid-container gap-y-0">
 		{
-			theme.Theme.images.length > 1 ? (
+			theme.Theme.images.length > 0 ? (
 				<ImageGallery
 					images={[theme.Theme.image, ...theme.Theme.images]}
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks for catching this!

PR #1477, opened 2025-03-24
New Featured Themes Page

TheOtterlord:
## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [ ] Safari
- [ ] iOS Safari


src/pages/themes/index.astro
```diff
@@ -0,0 +1,189 @@
...
+										{description}
+									</p>
+								</div>
+								<div class="relative">
+									<Image
+										src={image}
+										alt=""
+										class="rounded-lg xl:rounded-xl w-full h-48 object-cover"
+									/>
+									<div class="absolute inset-0 flex justify-center items-center opacity-0 group-hover:opacity-100 transition-opacity">
```
delucis, line 141: I think the hover styles look pretty good! Just a few notes:

- I think we need to dim the screenshots slightly with a dark overlay when displaying the “Browse…” call to action as contrast isn’t great currently (something like the theme cards on the homepage for example).
- We should make sure these hover styles also work when someone focuses the link. (Given we should move the link to the heading, this will likely need something like a group on the card, and `group-focus-within` or whatever it is in Tailwind?)
- I think it would be nice to always show the “Browse all themes” CTA on the last card, otherwise it feels a little empty.

src/pages/themes/index.astro
```diff
@@ -0,0 +1,189 @@
...
+									<h3 class="font-semibold text-white text-base">{title}</h3>
+									<p class="hidden sm:block font-light text-astro-gray-200 text-sm xl:text-base md:text-balance">
+										{description}
+									</p>
+								</div>
+								<div class="relative">
+									<Image
+										src={image}
+										alt=""
+										class="rounded-lg xl:rounded-xl w-full h-48 object-cover"
```
delucis, line 139: The fixed height here makes for a varying aspect ratio which I think causes some slightly awkward cropping. I think we can have a consistent aspect ratio with `w-full h-auto` probably given we control the size of the images on each of these.

src/pages/themes/index.astro
```diff
@@ -0,0 +1,189 @@
...
+
+					<h2
+						class="mt-1 text-white text-balance heading-4 sm:heading-3 md:heading-2"
+						id="categories"
+					>
+						Launch your next site at light speed with our catalog of free and premium Astro themes
+					</h2>
+				</div>
+
+				<div class="md:max-lg:hidden gap-2 lg:gap-4 grid grid-cols-1 lg:grid-cols-3 mt-4 sm:mt-8">
```
delucis, line 122: The gaps here are also pretty tight on larger screens — I think we could give the cards a bit more room to breathe.

src/pages/themes/index.astro
```diff
@@ -0,0 +1,189 @@
...
+										</div>
+									</div>
+								</div>
+							</a>
+						))
+					}
+				</div>
+			</section>
+		</div>
+
```
delucis, line 154: A few notes on the footer section. Here’s the preview (left) and design (right):

image

- Spacing should be evenly distributed above and below the horizontal border. Currently the border is skewed much closer to the Astro logo of the lower section in the deploy preview.
- The “Showcase your themes” headline should be larger at this breakpoint.
- The paragraph below the heading should use a regular sans-serif rather than Obviously.
- The button should use the larger variant.

Might be worth referencing [my similar component here](https://github.com/withastro/preview.astro.build/blob/chris/agency-pages/src/pages/agencies/_components/BecomeAPartnerCTA.astro), which you can probably copy/paste wholesale and just replace the content.

delucis (approved):
OK, should be good to merge when you’re ready @TheOtterlord! Thanks again for amazing work on this page. Great attention to detail 👏

PR #1497, opened 2025-04-02
Update to 18 themes per page

sarah11918:
Updating the number of themes shown per page.

This allows more themes to be discovered without the friction of navigating to a new page.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/pages/themes/[page].astro
+++ b/src/pages/themes/[page].astro
@@ -86,7 +86,7 @@ const allThemes: ThemeAndAuthor[] = await fetch(
 // take all matching themes and create a paginated list of results
 const paginatedResults = paginate({
 	data: allThemes.filter((theme) => !theme.Theme.featured),
-	pageSize: 12,
+	pageSize: 24,
 	currentPage,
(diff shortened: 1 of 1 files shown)


TheOtterlord (approved):

src/pages/themes/[page].astro
```diff
@@ -86,7 +86,7 @@ const allThemes: ThemeAndAuthor[] = await fetch(
 // take all matching themes and create a paginated list of results
 const paginatedResults = paginate({
 	data: allThemes.filter((theme) => !theme.Theme.featured),
-	pageSize: 12,
+	pageSize: 24,
```
Comment (delucis) [inline src/pages/themes/[page].astro:89]: I think we agreed on 18 rather than 24? The issue with really long pages is it can be a bit awkward scrolling up and down while the filter UI is still only up top.

```suggestion
	pageSize: 18,
```

delucis (approved):

PR #1514, opened 2025-04-08
Enable Expressive Code’s `frames` feature

delucis:
This PR enables Expressive Code’s `frames` feature, which includes file titles and copy code buttons.

I disabled it by default still for `sh`, `bash`, and `shell` code blocks to not add the terminal window bar (but this can still be added back with a `frame="terminal"` attribute if we ever need it).

In blog post code blocks that included a filename at the top of a snippet (e.g. `// astro.config.mjs`), these will now show that filename in the snippet title bar.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -22,7 +22,17 @@ export default defineConfig({
 		}),
 		astroExpressiveCode({
 			themes: [houston],
-			frames: false,
+			styleOverrides: {
+				borderRadius: '0.375rem',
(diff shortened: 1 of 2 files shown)


ematipico (approved):
Thank you! I checked the 5.6 blog post, and I can see the file names now

PR #1515, opened 2025-04-09
Fix code block styling on landing page

delucis:
We added some additional Expressive Code styles in #1514, but I missed that we were using it on the homepage too. @hippotastic spotted that the new styles clashed with the homepage usage. This PR fixes that.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


sarah11918 (approved):
Looks good on Chrome!

hippotastic (approved):
LGTM on both Chrome and macOS Safari.

PR #1536, opened 2025-04-20
Handle API 404 on theme details

TheOtterlord:
Handle 404s from portal to prevent the route from crashing if a theme does not exist.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


delucis (approved):
Thanks @TheOtterlord!

PR #1540, opened 2025-04-22
Add redirect for llms.txt

ascorbic:
Adds a redirect from `/llms.txt` to the docs site.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -22,6 +22,7 @@
 
 # Documentation shortcuts
 /docs/                  https://docs.astro.build/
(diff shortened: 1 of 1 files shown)


florian-lefebvre (commented):
1. Do LLMs handle redirects? If not we may need to make it a rewrite instead
2. There are 3 files to redirect, see https://delucis.github.io/starlight-llms-txt/

florian-lefebvre (approved):

PR #1543, opened 2025-04-24
Remove `width="auto"` attribute from sponsor icons

delucis:
This PR fixes a small “bug” in the markup for our sponsors grid. When updating to Astro Icon to v1, I fudged the sponsor grid layout using `width="auto"` because v1 of Astro Icon _always_ includes a `width` attribute which broke our responsive layout. `width="auto"` is invalid HTML, but it did fix things because it stopped Astro Icon adding a valid `width` attribute.

This PR fixes it properly by allowing Astro Icon to add its `width` attribute and then overriding it with `width: auto` CSS.

Not a big deal, but this “bug” did call console warnings so wanted to get it cleaned up.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/pages/_components/landing-page/Sponsors.astro
+++ b/src/pages/_components/landing-page/Sponsors.astro
@@ -96,9 +96,8 @@ const sponsors = [
 						class="h-40 p-4 size-full basis-1/3 flex flex-col lg:grid lg:grid-rows-subgrid lg:row-span-full items-center justify-center gap-4 -outline-offset-2 hover:bg-astro-gray-100/5 bg-clip-padding transition-colors duration-400"
 					>
 						<Icon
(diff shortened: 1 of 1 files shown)


sarah11918:
This is how it looks to me on Chrome desktop: logos are tiny?

image

delucis:
Wait, what, I tested in Firefox and it looked fine, but I see that now in FF too 😅 

Thanks for testing @sarah11918!

delucis:
OK, pushed up an alternative fix. Bit of a hack, but seems to be the only way to get the behaviour I want from Astro Icon.

ematipico (approved):
Tested on Firefox mobile and desktop and they look good

PR #1558, opened 2025-05-01
Fix blog post OG image

delucis:
Fixes the OG image for the April what’s new blog post.


HiDeoo (approved):
Looks good to me 👍 

image

PR #1644, opened 2025-06-27
fix: astro db command in Deep Dive

trueberryless:
I found a small typo in a command and I think it's worth fixing it alto this post is rather not one of the newest ones 😅


delucis (approved):
Thanks @trueberryless!

PR #1685, opened 2025-07-12
feat: add Starlight's logomarks and adapt the grid

trueberryless:
I've been using the Astro Press page frequently to access logo resources and noticed a small inconsistency: while the Astro logos provide separate icons, the Starlight logos are only available as full lockups. Since the Starlight icon (both in dark and light variants) can be easily extracted from the full logo, it makes sense to offer these standalone icons directly on the Press page as well.

For consistency, I’ve matched the icon dimensions (`107px` height) to those of the existing Astro icons, so they receive equal visual weight and importance on the page.

I believe this addition would make the Press page more complete and user-friendly for designers and developers alike.


trueberryless:
This seems like a GH bug, in light mode those black areas do not show up:
image

However, when taking a closer look at the images, I see that the bottom seems to be cut off a little bit, so I am going to look if I can improve the images, thanks for taking a closer look!
image

PR #1705, opened 2025-07-22
Promos

delucis:
Experimental PR playing with the idea of cross promoting some of our content in our blog pages.

How this works:

- This PR adds a new `promos` content collection, which is a YAML file defining promotions and some images.
- On individual blog posts a promo is added at the bottom of the post content. On wide enough screens this is moved to a sticky position in the bottom right corner while readers scroll.
- Promos are also added to the side panel on the main `/blog/` index.

Currently there are two promos defined — mainly to test out the implementation with varying content. We can discuss what we want to promote currently and how to manage this going forward.

One additional consideration is whether we want individual blog posts to be able to declare a preference for the promo to display, e.g. perhaps we feel a specific promo is particularly relevant to a blog post’s audience and wouldn’t want to just randomize the displayed promo. We could implement something like that if there’s interest.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


sarah11918:
This looks amazing, Chris! The hover animation is smooth and subtle.

- Not sure whether we care, but on Chrome desktop, scrolling down to the footer, at *some* point, the positioning stops being sticky in the viewport and stays locked to not quite the end of the blog post content vs the end of the scrollable content (related posts and footer):

	image

- Similarly, when at the very top of the page, it *might* be preferable if the promo weren't half-shown (kicks in for me at 1536px)? (I like the idea of having it not visible until any scrolling down, to make a blog post title/author the only thing visible? Otherwise, I feel like it might make more sense to see the whole thing.)

	image

- Just an aside: this is making me realize that we should maybe be adding a `## Related posts` heading in the blog posts when we include them? 😄 I don't really like the list of names running directly into cards with no separation/indication. I was even thinking, should the promo come *before* them, but remembered they are part of the content, and so probably like every other section in the blog post, should have a heading.

	image

HiDeoo:
I personally love the idea and implementation, great use of container queries (and asides :smile:), and the overall design is lovely and functional while not being obtrusive at all.

I had a few positioning questions but that's covered by Sarah's comment and screenshots.

On the `/blog/` index, I personally wonder if it should not even be above the RSS feed link. When RSS was very popular, it was already difficult to find them, it was usually at the bottom of the page, or you had to rely on the browser to detect it or find it in the source code, so back to now, having the promo before it would not be a bad idea imo and potentially make it more visible to users.

delucis:
Thanks both:

- Addressed the top/bottom alignment observations by tweaking how the sticky positioning is implemented. We can’t guarantee the promo will never show “above the fold”, but it does start a little lower down now, and it sticks to the bottom of the content column rather than being offset.
- I moved the promo above the RSS subscribe panel in the blog index pages as @HiDeoo suggested.
- I also fixed the horizontal position of the sticky promo to accommodate the showcase grids used in “What’s new” blog posts, which break out beyond the default column width.

> Just an aside: this is making me realize that we should maybe be adding a ## Related posts heading in the blog posts when we include them?

Now that we’ve settled on liking these, we can maybe switch to implementing related posts via frontmatter rather than the ad-hoc inclusion of components in every post. That could give us the freedom to include them after promos. Could do that in a follow-up PR.

sarah11918 (approved):

delucis:
> Can we track the clicks to see if it works?

Good shout — added in 6a35c7bca4279bea881b8d7064390974d991bbe5

PR #1715, opened 2025-07-28
Move related blog posts functionality to frontmatter

delucis:
This PR introduces a new `related` frontmatter property for blog posts and drops our manual `<RelatedPosts>` component in favour of this new property.

Now, instead of including `<RelatedPosts slugs={[...]} />` in a blog post, you can add them as frontmatter properties instead.

This allows us to more easily separate these “footer” elements in the flow, for example, placing the [recently added promos](https://github.com/withastro/astro.build/pull/1705) above these links.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


src/pages/blog/[slug].astro
```diff
@@ -117,15 +118,27 @@ const structuredData = {
...
+				>
+					<div class="2xl:ml-[9rem] 2xl:w-56 2xl:sticky 2xl:bottom-12">
 						<Promo />
 					</div>
 				</div>
+				{
+					data.related.length > 0 && (
+						<hr class="-mx-16 basis-[calc(100%+8rem)] border-astro-gray-500" />
+						<nav aria-labelledby="_related-posts">
+							<h2 class="heading-3" id="_related-posts">Previously…</h2>
```
Comment (sarah11918) [inline src/pages/blog/[slug].astro]: I do wonder whether "Related Posts" makes more sense as a heading, given that these won't *always* be releases in a sequential order? e.g. If we have a more community type of post, I'm not sure "Previously... how Astro does i18n" *exactly* fits vs just saying, "Here's something sort of random but in that ballpark" -- like the Community Loaders and the specific loader blog posts that weren't *actually* previous, but released at the same time?

```suggestion
							<h2 class="heading-3" id="_related-posts">Related Posts</h2>
```
Don't feel super strongly, just pointing it out!

src/pages/blog/[slug].astro
```diff
@@ -117,15 +118,27 @@ const structuredData = {
...
-					<div class="my-12 2xl:ml-[9rem] 2xl:w-56 2xl:sticky 2xl:bottom-12">
+				<hr class="2xl:hidden -mx-16 basis-[calc(100%+8rem)] border-astro-gray-500" />
+				<div
+					class="w-full 2xl:w-0 2xl:mt-24 flex flex-col 2xl:before:h-full"
+				>
+					<div class="2xl:ml-[9rem] 2xl:w-56 2xl:sticky 2xl:bottom-12">
 						<Promo />
 					</div>
 				</div>
+				{
```
Comment (sarah11918) [inline src/pages/blog/[slug].astro:133]: Just a comparison of the styling (gaps, where the promo sits) on my 24 inch monitor new vs old. Seems like this PR introduces more vertical space? (Not sure it matters where the promo actually sticks, but worth noting it now seems to stick to content vs the related posts.)

This PR:
image

Currently:
image

(And, I was actually surprised that at 1026px, the layout goes from two blog posts in a row to one. I'm usually looking at this on my Chromebook, where the two related posts are side by side. Is there a reason on BIG screens we go back to only one in a row?)

sarah11918 (commented):
Thanks Chris! I think this is much nicer for authring the posts! Just some comments below!

src/pages/blog/[slug].astro
Comment (delucis) [reply src/pages/blog/[slug].astro:133]: > I was actually surprised that at 1026px, the layout goes from two blog posts in a row to one

Chromebook, so Chrome obviously. Very odd — not something I see but could it be an old version of Chrome in your case?

src/pages/blog/[slug].astro
Comment (delucis) [reply src/pages/blog/[slug].astro]: Yeah, I mostly just copied what we had used so far. Do you think “Related Posts” is the correct user-facing title? Or do you think you tended to associate “related” with this because that’s what I’ve called the feature and frontmatter property?

Comment (sarah11918) [reply src/pages/blog/[slug].astro]: I think "Related posts" is a pretty standard concept (someone literally just posted an article they found about "Smarter related posts in Astro" in showcase), so I do think it's probably correct (even if not the only possibility). These are *posts*, and we've tagged them because they're somehow *related*. Seems to fit to me!

Comment (delucis) [reply src/pages/blog/[slug].astro]: Perfect, thank you!

src/pages/blog/[slug].astro
Comment (sarah11918) [reply src/pages/blog/[slug].astro:133]: This is my Linux Chrome desktop. (And I literally updated Chrome on Friday). So, I don't *think* it's an outdated version? Version 138.0.7204.168 (Official Build) (64-bit)

(Confirmed, this is the latest version of Chrome for Linux)

Comment (sarah11918) [reply src/pages/blog/[slug].astro:133]: On my Chromebook, it threatens to show a larger size with two posts in the row. So, it's just Linux (the shots above).

Chromebook inspected:

image

Comment (delucis) [reply src/pages/blog/[slug].astro:133]: Yeah, up to date alright. Very weird — I can’t see this on any browser, including Chrome, on macOS, so I guess it’s something specific to the Linux version.

Comment (sarah11918) [reply src/pages/blog/[slug].astro:133]: Confirming the 14" Android mobile Chrome is showing the two posts in a row, too. It just seems to be Linux for me.

Comment (sarah11918) [reply src/pages/blog/[slug].astro:133]: Just reporting that if I inspect, there appears to be a media query (which I don't think we're using?) and indeed, removing this fixes it. 

An adventure for another day, perhaps!

image

sarah11918 (approved):
The latest commit even fixes the existing Linux/Chrome desktop layout issue! :tada:

PR #1723, opened 2025-07-31
Downgrade Ubuntu in showcase update workflow

delucis:
This PR downgrades the version of Ubuntu used to run our showcase update workflow.

#1611 updated this, but it’s causing Puppeteer to fail in that workflow: https://github.com/withastro/astro.build/actions/runs/16646443672/job/47108742183


sarah11918 (approved):
Thank you!

PR #1736, opened 2025-08-05
August sponsor update

delucis:
Updates our homepage links and repo image for August.


public/sponsors.png
ascorbic, line 1: Should probably update Firebase on this one too

delucis: Oh yeah, missed we’d done that elsewhere but not here — good catch. Done in a7d639745ca3fc95d2b9853686f2b18aed9467bd.

PR #1741, opened 2025-08-06
Feature cards update

delucis:
This PR updates the “Fully Featured” section of the astro.build homepage to include a few more features and add illustrations.

It also improves the animations these cards have on hover by making them work on keyboard focus as well, plus on viewport-enter on devices that don’t support hover. (All while respecting reduced motion preferences.)

Another a11y improvement: these cards previously were implemented as a link containing all the text, making for very verbose link labels. Now only card titles are used as links while still keeping the full card clickable.

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [x] Android Firefox
- [x] Safari
- [ ] iOS Safari


sarah11918:
Looks fantastic! I'll review the card text specifically, but two quick initial comments on the visuals are:

- hahah, was not expecting the "unhover" animation on the components 😂 
- feels like maybe the dev toolbar card at the bottom could be another candidate for a visual, since the toolbar *is* visual, and since it's at the end and there hasn't been an image for a few cards?

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
```diff
@@ -0,0 +1,183 @@
+---
+import FeatureCard from './FeatureCard.astro';
+---
+
+<FeatureCard
+	title="Optimized Images"
+	description="Eliminate layout shift and serve highly optimized, modern image formats with Astro’s built-in components and utilities."
```
sarah11918: Should we mention the word "responsive" in here somewhere, maybe?

src/pages/_components/landing-page/feature-cards/ServerTemplating.astro
```diff
@@ -0,0 +1,116 @@
+---
+import FeatureCard from './FeatureCard.astro';
+---
+
+<FeatureCard
+	title="Simple Templating"
+	description="If you know HTML, you already know enough to write your first Astro component. Build powerful server components everyone on your team understands."
```
sarah11918, line 7: Does the phrase "server components" make people think of "React server components", and is that a good thing or a bad thing? 😅 

I'm realizing we don't tend to say "server components" in the docs, but define it negatively as no client runtime, don't render on the client, etc. Not opposed, just noting that we don't otherwise use this phrase commonly ourselves, I don't think?

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
HiDeoo, line 1: The `prefers-reduced-motion` support of this card seems a bit off compared to the other card. I wonder if an approach similar to the one used in `FileBasedRouting.astro` should be used where the default version is the end, and if `prefers-reduced-motion: no-preference` is detected, we apply the transform for the final state and remove the transforms on focus/hover? That would prevent any movement like the other cards.

src/pages/_components/landing-page/feature-cards/FeatureCard.astro
```diff
@@ -0,0 +1,90 @@
...
+	/** Toggle the `in-viewport` class name on an element. */
+	const markVisible = (el: Element, isVisible: boolean) =>
+		el.classList.toggle('in-viewport', isVisible);
+
+	const observer = new IntersectionObserver(
+		(entries) => {
+			for (const entry of entries) markVisible(entry.target, entry.isIntersecting);
+		},
+		{ threshold: 0.8 },
+	);
```
HiDeoo: While this feel great when a card enter the viewport, e.g. "oh, this card just started animating right after what I'm currently focusing on, I should take a look", I feel like reverting an animation should happen when it's really offscreen.

For example, if the "Optimized image" card has played its animation because it's the viewport, I feel the exit animation being triggered at the top of my screen while I'm scrolling away, and supposedly focusing on the next content, make me focus back on the top on a content I already passed/read.

src/pages/_components/landing-page/feature-cards/ServerTemplating.astro
HiDeoo, line 1: I think one key difference between this animation and all the other ones, is that part of it covers the text. It feels a bit "heavy", specially for the exit animation, where you can end up in cases like this:

image

I wonder if a smaller scale change, maybe a play on opacity during the transition, or maybe a slightly smaller duration coup help. Or maybe only tweaking the exit animation, not sure 🤔

HiDeoo (commented):
What an amazing PR 🤩 Such lovely cards 👏

I left a few comments but as we know, when it comes to UI, I'm not the best to make decision so feel free to discard.

src/pages/_components/landing-page/feature-cards/ServerTemplating.astro
delucis: Yeah, uncommon for us, but chosen intentionally for marketing reasons 😅

ArmandPhilippot:
This looks really great! 🙌🏽 

But I wonder, is the different color on focus using keyboard navigation is intentional? Most of the other focusable elements use a blue color while these features cards use a pink color?
Photo montage of different parts of Astro.build UI when focused and showing a different color

Actually, I do like the change, it looks smoother on my end! Just wondering if the use of a different color could be confusing?

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
delucis: I think I even did in an earlier iteration! Can you see somewhere it would fit? I kind of feel like this one’s reached its adjective limit.

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
delucis: Ah yeah, I wondered if the state change without a transition would be OK? i.e. that it’s the transition of objects moving in distinct directions that can trigger issues with vestibular disorders etc. rather than any visual change. But if you think it would be better we can definitely do as you suggest.

HiDeoo: Technically I think it's fine, I only mentioned it compared to the other cards, altho, I was definitely focusing on the cards during the review so it may just be a detail that we can ignore for now until we get some feedback on it. Still wanted to mention it tho.

src/pages/_components/landing-page/feature-cards/ServerTemplating.astro
delucis: Tried the opacity change suggestion and I quite like it (pushed in [f2cc4d4](https://github.com/withastro/astro.build/pull/1741/commits/f2cc4d4aeabae047c82437d5113b0491f2f32f90)). What do you think?

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
sarah11918: If we cared about getting "responsive" in there, could probably streamline:

```suggestion
	description="Eliminate layout shift and serve optimized, responsive images with Astro’s built-in components and utilities."
```

- is "highly optimized" like "very unique"? Maybe this word could be cut?

- perhaps "modern formats" isn't a super flex, because people can create whatever images they want in whatever format they need themselves before uploading? (And, maybe still not all browsers support all formats?) Maybe easy *responsive* image behavior is something that more addresses a pain point they have/recognize?

src/pages/_components/landing-page/feature-cards/ServerTemplating.astro
HiDeoo: Oh, I think it looks way better, and I also think this may be enough to entirely fix what I try to describe, no other changes necessary imo 👍

src/pages/_components/landing-page/feature-cards/FeatureCard.astro
delucis: Great point. Pushed 17fb11f7661aaf52a21b08f8efa3631532910c2f for this. It uses two intersection observers: one to apply the `in-viewport` class once the card is 80% visible as before, and a new one to remove the class only once the card is 100% out of the viewport. Felt OK in my tests but would love another look.

HiDeoo: Tested on mobile, tablet, and even desktop with touch mode enabled, and it looks perfect to me 👌

delucis: Awesome, thanks again for the good feedback 🙌

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
delucis: Hmm, I actually think modern formats trumps “responsive”. Converting to WebP and co. is not the simplest — JPEG, PNG, etc. are far more common if you just have photos or are exporting from graphics software — and browser support is good. (And for this audience I think “modern” also sounds good more than threatening/risky 😄)

I think maybe best to leave it. But we can remove the “highly” 👍 

```suggestion
	description="Eliminate layout shift and serve optimized, modern image formats with Astro’s built-in components and utilities."
```

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
delucis: Pushed an improvement on this one where we can still transition but just use the opacity/blur filters for users requesting reduced-motion — 17fb11f7661aaf52a21b08f8efa3631532910c2f. Feels much nicer!

ArmandPhilippot:
> Hmm, seems to be an existing Firefox vs Chrome difference 🤔 In Chrome all elements have the pink outline on focus, but seems like only some do in FF for some reason.

Oh interesting, I hadn't tested with Chromium. But I think my devices are as evil than Sarah's ones... I also see a difference in Chromium. 😅 Where Firefox uses some blue, Chromium uses white!? And featured cards uses pink in both.

While, in the live website, everything uses white in Chromium (and blue in Firefox) except the "Get started" button that uses pink.

src/pages/_components/landing-page/feature-cards/ImageOptimization.astro
HiDeoo: Amazing, feels really great and more aligned with the other cards 🙌 (that's [`6daf196` (#1741)](https://github.com/withastro/astro.build/pull/1741/commits/6daf1963751c74046ecccdfcd2df91b9262c2cce) tho)

src/pages/_components/landing-page/feature-cards/ServerTemplating.astro
bjohansebas, line 1: I think it would be great if we respected the users' prefers-reduced-motion setting. It's already being done in several of the other cards that have complex animations, so we can do it in this one too

HiDeoo: It seems that it's also respected in this card.

Without any preference:

https://github.com/user-attachments/assets/3a2dc38c-70a7-47dc-80f4-6f90a92ae779


With prefers-reduced-motion:

https://github.com/user-attachments/assets/cdfb77c3-1b1a-4e12-ae27-fde347ba0252

Is this not what you are experiencing? Or do you mean it should remove more animations vs only the non-essential motions that usually trigger disconfort for users with vestibular motion disorders?

bjohansebas: hadn't noticed much of a difference until I tried it again, but I think it's fine like this

OliverSpeir (approved):
Looks great on iOS

PR #1742, opened 2025-08-07
Tweak ecosystem theme cards

HiDeoo:
As we talked about it this Monday, this PR initially aims to tweak the gradient visible when hovering a theme card on the landing page which is a bit "abrupt", specially visible on cards with a white image. Altho, this PR includes a few other changes:

- Refactor the ecosystem cards into a component to avoid code duplication and fixing things in many components. A new `<EcosystemPanel>` component has been created and it accepts an extra `moreCard` or `moreLink` prop to display the "more" card or link.
- For the "Landing Pages" tab, the more card use the same title and link as the "Trending" tab, so I fixed that.
- As planned, I slightly tweaked the gradient on hover to be more subtle (not changed for the more card as it's already a different gradient which seems fine in this case).
- The "View theme" link on cards had an hover effect even tho the entire card is clickable, which could be confusing considering clicking the "View theme" area does not have a different effect than clicking the card itself. I removed the hover effect and applied it by default when hovering the card.
- I slightly increased the background opacity of the "View theme" link to make it more visible on some cards as I had trouble reading it on some cards.
- The more card was not entirely clickable and only its link was. Maybe this was a design choice, but as I found it weird multiple times while working on this PR, I made the entire card clickable but we can revert it if needed.

I'm perfectly fine with reverting any of these changes if some of them are not desired. I also found some other potential improvements in this area of the landing page while working on it, but I decided that this PR was already large enough with some undiscussed changes.

| Before | After |
| --- | --- |
| SCR-20250807-ljvq | SCR-20250807-lsod |

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [x] Safari
- [x] iOS Safari


src/pages/_components/landing-page/EcosystemPanel.astro
```diff
@@ -0,0 +1,63 @@
...
+<div class="p-1 grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 gap-4 rounded-2xl">
+	{themes.map((theme) => <ThemeContainer theme={theme} />)}
+	{
+		moreCard && (
+			<div class="size-fit group relative">
+				<Image
+					width={800}
+					class="width-full rounded-xl isolate select-none"
+					src={moreCard.image}
+					alt="An image showing all types of content driven websites you can build with Astro"
```
delucis: What do you think about removing this?
```suggestion
					alt=""
```
I spotted it the other day and I don’t _think_ it’s at all helpful? The image seems to be decorative to me.

src/pages/_components/landing-page/ecosystem-tabs/ThemeContainer.astro
```diff
@@ -31,10 +31,10 @@ const { theme } = Astro.props;
 		alt={`${theme.title} theme by ${theme.author}`}
 	/>
 	<div
-		class="m-auto size-full hidden group-hover:landing-section absolute inset-0 z-10 bg-gradient-to-b from-transparent to-30% to-astro-dark-900/80 rounded-xl"
+		class="m-auto size-full hidden group-hover:landing-section absolute inset-0 z-10 bg-gradient-to-b from-astro-dark-900/50 to-90% to-astro-dark-900/90 rounded-xl"
 	>
 		<div
-			class="px-4 py-2 inline-flex items-center justify-center gap-2 rounded-full border text-base font-normal transition-all ease-in-out duration-500 hover:cursor-pointer font-sans font-bold border-astro-dark-100/30 hover:border-astro-dark-100/60 text-astro-gray-200 hover:text-astro-gray-100 bg-astro-dark-600/30 hover:bg-astro-dark-600/30"
+			class="px-4 py-2 inline-flex items-center justify-center gap-2 rounded-full border text-base font-normal transition-all ease-in-out duration-500 font-sans font-bold border-astro-dark-100/60 text-astro-gray-100 bg-astro-dark-600/50"
```
delucis: Noticing that this change removes the style changes on hover while still keeping `transition` and related classes, I guess maybe we should clean those up?

Although I also quite liked the way the button itself brightened up previously. Was there a reason you decided to remove that aspect? If we remove it, maybe a subtle transition on the gradient opacity could be a good way to keep the interaction feeling slightly “softer”.

delucis (commented):
Lovely work! Always happy to see code polished up into a reusable component — feels so clean! 🧼 

Left a couple of small notes but overall it’s looking amazing 👏

src/pages/_components/landing-page/ecosystem-tabs/ThemeContainer.astro
HiDeoo: The `transition` and related classes is something i missed indeed as my intention was to remove that change entirely.

My reasoning for the change was this point:

> The "View theme" link on cards had an hover effect even tho the entire card is clickable, which could be confusing considering clicking the "View theme" area does not have a different effect than clicking the card itself. I removed the hover effect and applied it by default when hovering the card.

It may not be clear but basically:

- The card is visible, I'm not hovering it, the "View theme" "button" is not visible.
- I hover the card, the "View theme" "button" is visible, the gradient appears, my cursor changes, and I'm on a link.
- I now move my cursor over the "View theme" button, the button gets different styles applied kinda like it's a different element/behavior if I were to click on it vs clicking the card itself.

This is the part that I personally find confusing and wanted to avoid. I guess we can:

- It's actually fine and it's just me not liking it. We revert this change.
- We keep it, remove the extra `transition` classes I forgot, and add a transition on the gradient opacity like you suggested.

Let me know what you think and prefer, I'll make the changes accordingly.

src/pages/_components/landing-page/EcosystemPanel.astro
HiDeoo: This was part of the "other potential improvements in this area of the landing page" I mentioned so happy to go with it, thanks for suggesting it 🙌

src/pages/_components/landing-page/ecosystem-tabs/ThemeContainer.astro
delucis: That makes sense. Definitely agree that whatever hover styles we agreed on, should apply when the card is hovered, not just the button.

I think probably just fading in the full overlay (scrim + button) with an opacity transition is sufficient 👍

HiDeoo: Just pushed an update adding an opacity transition, let me know if this is what you had in mind.

src/pages/_components/landing-page/ecosystem-tabs/ThemeContainer.astro
```diff
@@ -31,10 +31,10 @@ const { theme } = Astro.props;
 		alt={`${theme.title} theme by ${theme.author}`}
 	/>
 	<div
-		class="m-auto size-full hidden group-hover:landing-section absolute inset-0 z-10 bg-gradient-to-b from-transparent to-30% to-astro-dark-900/80 rounded-xl"
+		class="m-auto size-full landing-section absolute inset-0 z-10 bg-gradient-to-b from-astro-dark-900/50 to-90% to-astro-dark-900/90 rounded-xl opacity-0 transition-opacity ease-in-out duration-500 group-hover:opacity-100"
```
delucis: Nit, but I might go with just a 300ms transition on these. It’s true that a lot of hover interactions on the page are 500ms, but these feel slightly on the sluggish side given we’re revealing important extra context to users about what a click will do.
```suggestion
		class="m-auto size-full landing-section absolute inset-0 z-10 bg-gradient-to-b from-astro-dark-900/50 to-90% to-astro-dark-900/90 rounded-xl opacity-0 transition-opacity ease-in-out duration-300 group-hover:opacity-100"
```

src/pages/_components/landing-page/EcosystemPanel.astro
```diff
@@ -0,0 +1,63 @@
...
+							'px-4 py-2 inline-flex items-center justify-center gap-2 rounded-full border text-base font-normal font-sans font-bold border-astro-dark-100/30 group-hover:border-astro-dark-100/60 text-astro-gray-200 group-hover:text-astro-gray-100 bg-astro-dark-600/50',
+							'after:absolute after:inset-0 after:z-10 supports-[selector(:has(:focus-visible))]:focus-visible:outline-none',
+						]}
+					>
+						<span>{moreCard.title}</span>
+						<Icon name="ri/arrow-right-line" size={18} />
+					</a>
+				</div>
+			</div>
+		)
```
delucis: Looks like the more card would also need the same treatment.

HiDeoo: Definitely, I wanted to make sure we agree on the transition first tho as the same opacity transition would probably not be suited here as the overlay and button are visible by default, and if I remember correctly, you cannot animate gradient background so will need to figure out another way to do so 👍

HiDeoo: Cheated a little bit in a first attempt: I transition colors only on hover, which means:

- For the overlay, transition to a non gradient version of the background
- For the button, transition border and text color

delucis (approved):
> Cheated a little bit in a first attempt: I transition colors only on hover

Oh, this looks great — awesome work 👏 

I think this is ready to go. Thanks again for the attention to detail. Definitely a great improvement 👨‍🍳💋

PR #1755, opened 2025-08-14
chore: update teaclient website in showcase

Eveeifyeve:
## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


trueberryless (approved):
The changes per se look good to me.

However, it seems like this website is still rather work in progress, so I wonder if it should already be included in the showcase. Because I am not sure if showcase image will be updated automatically when the real content will be available. So we might need to update it manually as well when it's ready. Would you do that if the time comes please?

delucis (approved):
Thanks for the update @Eveeifyeve!

@ematipico @trueberryless This PR is only updating the URL of an existing showcase site to the correct one — this site is already in the showcase!

It’s true that the screenshot won’t update automatically, but folks are always welcome to PR screenshot updates if things change.

PR #1783, opened 2025-08-23
Landing page performance optimizations

delucis:
## Notes

- Various landing page image optimizations (lazy loading, width/height attributes, actually optimizing some that weren’t)
- Adds a missing recommended `initial-scale` meta value to our layout

- I also experimented briefly with `content-visibility` as a way to reduce layout computation, but that’s going to need more testing to get right and this PR has already been sitting for ages, so I reverted that experiment to get this out.
	- Something to consider if we do want to use [`content-visibility`](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Using_CSS_containment#content-visibility) one day: https://infrequently.org/2020/12/resize-resilient-deferred-rendering/

## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [x] Safari
- [ ] iOS Safari

### DIFF
--- a/src/pages/_components/landing-page/HeroBackground.astro
+++ b/src/pages/_components/landing-page/HeroBackground.astro
@@ -8,7 +8,6 @@ import heroBackground from '../landing-page/_assets/HeroBackground.png';
 	quality={10}
 	width={256}
 	priority
-	loading="eager"
 	alt=""
 	class="blur-lg absolute -z-50 -translate-x-1/2 inset-x-1/2 -inset-y-full w-[2353px] h-[1969px] object-cover max-w-[unset]"
 />


src/pages/_components/landing-page/HeroBackground.astro
```diff
@@ -1,14 +1,14 @@
...
-const background = await getImage({ src: heroBackground, quality: 10, width: 256 });
 ---
 
-<img
+<Image
+	src={heroBackground}
+	quality={10}
+	width={256}
+	priority
 	loading="eager"
```
HiDeoo: Nit: Isn't this [set by `priority`](https://docs.astro.build/en/reference/modules/astro-assets/#priority)?

```suggestion
```

delucis: Ah good catch! Yes, it’s handled automatically by `priority`.

PR #1880, opened 2025-10-07
Remove dead site from showcase

psncl:
I was browsing the [showcase](https://astro.build/showcase/) a few hours and saw that [Designcember](https://designcember.com/) shows there as the second website on the very first page. But it is a dead site now. The domain is still active, but looks like the website no longer exists. I have uploaded a screenshot [here](https://github.com/user-attachments/assets/c732a55d-7a85-40df-b6f3-8e2e54d8ae0b).

I have removed the corresponding `webp` and `yml` file for that site from the `src\content\showcase` folder. I could not find any other reference to `designcember` in the codebase, so I am assuming removing those two files is all that needs to be done.

I am new to open source and this is my first time making a pull request, so please let me know if I made any mistakes.


delucis (approved):
Oh, too bad it’s this site, I always liked it 😄 

I’ve opened an issue in their repo just to double check it’s gone for good — https://github.com/GoogleChromeLabs/designcember/issues/64 — but your PR looks perfect for removing it, thank you!

PR #1926, opened 2025-10-27
Fix ID used in landing pages themes tab

### LINKED ISSUE #1925: Webaim detecting
I decide to test with WebAIM and found a small number but one confusing is "Broken ARIA reference".

https://wave.webaim.org

delucis:
Closes #1925 

This PR fixes the ID used to associate the “Landing pages” theme tab panel with the tab. This currently uses `aria-labelledby` to associate them, but the implementation was using an ID containing a space.

While changing, I also cleaned up use of `data-tab` attributes, which are not used for anything currently and so are superfluous.

## Browser Test Checklist

- [ ] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/src/pages/_components/landing-page/EcosystemCategory.astro
+++ b/src/pages/_components/landing-page/EcosystemCategory.astro
@@ -9,7 +9,6 @@ const { id, tab, class: className } = Astro.props;
 ---
 
 <section
-	data-tab={tab}
 	role="tabpanel"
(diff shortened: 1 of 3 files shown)


HiDeoo (approved):
Great catch 🕵️ Checked the preview to confirm the fix and looks perfect to me.

PR #1964, opened 2025-11-10
Migrate content collections to all use loaders

delucis:
This PR updates our content collection configuration to migrate all collections to use loaders, moving the config to `src/content.config.ts`, and updating collection usage to match.

Might need some careful checking to make sure I haven’t missed somewhere causing e.g. broken links.

## Browser Test Checklist

- [x] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


delucis:
YOLO! Let’s roll with it 🚀

PR #2060, opened 2025-12-18
Astro v6 beta test

delucis:
Draft PR to play around with the v6 ~~alpha~~ beta to see what happens


package.json
```diff
@@ -21,18 +21,18 @@
 	},
 	"dependencies": {
 		"@astro-community/astro-embed-youtube": "^0.5.9",
-		"@astrojs/check": "^0.9.6",
-		"@astrojs/mdx": "^4.3.13",
-		"@astrojs/netlify": "^6.6.3",
-		"@astrojs/rss": "^4.0.14",
-		"@astrojs/sitemap": "^3.6.0",
+		"@astrojs/check": "0.9.6-alpha.0",
```
alexanderniebuhr: Actually `0.9.6-alpha.0` is older than `0.9.6`. I think that is a bug in the upgrade command 🤔

delucis: I _think_ that’s expected? It just means we published a stable release (from `main`) since first creating the alpha?

.gitignore
```diff
@@ -32,3 +32,5 @@ pnpm-debug.log*
 
 # Local Vercel folder
 .vercel
+
+.idea
```
delucis: Not sure where this slipped in:

```suggestion
```

delucis:
@matthewp Before we merge, I’m going to do some diffs of the site HTML output as I haven’t done that yet. Everything _looks_ right in the browser, but probably worth an inspection.

PR #2108, opened 2026-01-16
fix: Incorrect Sentry URL

timfish:
The URL for Sentry was wrong.


sarah11918 (approved):
Thanks for the fix! :rocket:

PR #2139, opened 2026-01-29
Fix `update:showcase` script

HiDeoo:
This PR fixes the way we import `@actions/core` in the showcase update script following the [`3.0.0` release](https://github.com/actions/toolkit/blob/main/packages/core/RELEASES.md#300) of the package.

## Test Checklist

I tested locally, before the fix we get:

```
import ghActions from '@actions/core';
       ^^^^^^^^^
SyntaxError: The requested module '@actions/core' does not provide an export named 'default'
```

After the fix, the script no longer error and starts downloading Chrome (note that I did not let it run entirely locally).

## Browser Test Checklist

- [ ] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

### DIFF
--- a/scripts/update-showcase.mjs
+++ b/scripts/update-showcase.mjs
@@ -1,5 +1,5 @@
 import fs from 'node:fs/promises';
-import ghActions from '@actions/core';
+import * as ghActions from '@actions/core';
 import octokit from '@octokit/graphql';
 import { parseHTML } from 'linkedom';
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Approving as this seems consistent with the error, and the run gets further than I was getting! Thank you!

PR #2376, opened 2026-05-04
feat: Astro 6.3 release article

matthewp:
## Summary
- Draft release article for Astro 6.3 (shipping May 7)
- Advanced Routing section ported from the removed 6.2 draft
- Placeholder sections for image redirect support, resilient island hydration, and SVG flag gating
- Cover and OG images converted to webp/jpg conventions


src/content/blog/astro-630.mdx
```diff
@@ -0,0 +1,196 @@
...
+export default defineConfig({
+  experimental: {
+    advancedRouting: true
+  }
+});
+```
+
+### Quick start with `astro()`
+
+Create a `src/app.ts` file that uses the combined `astro()` handler to run the full Astro pipeline:
```
ematipico: Maybe we should add a note/warning at the end of advance routing presentation, and say that `src/app.ts` becomes a known file in Astro, so if users already have `src/app.ts` in their code, their app might break

src/content/blog/astro-630.mdx
```diff
@@ -0,0 +1,196 @@
...
+    dangerouslyProcessSVG: true,
+  }
+});
+```
+
+This change does not affect importing SVGs as components. It only applies to rasterizing SVG sources through the image optimization pipeline (e.g. converting an SVG to PNG via `<Image />`).
+
+## Other improvements
+
+- **`consume()` method for `AstroCookies`**: A new `consume()` instance method on `AstroCookies` marks cookies as consumed and returns the `Set-Cookie` header values. After consumption, any subsequent `set()` calls will log a warning since the headers have already been sent. This replaces the static `AstroCookies.consume(cookies)` method, which is now deprecated but kept for backward compatibility with existing adapters.
```
ArmandPhilippot: nit: I feel that the repetition makes reading a little awkward.
```suggestion
- **A new `consume()` method for `AstroCookies`**: this method marks cookies as consumed and returns the `Set-Cookie` header values. After consumption, any subsequent `set()` calls will log a warning since the headers have already been sent. This replaces the static `AstroCookies.consume(cookies)` method, which is now deprecated but kept for backward compatibility with existing adapters.
```

ArmandPhilippot (approved):
A nit for readability, otherwise LGTM! 🚀

src/content/blog/astro-630.mdx
matthewp: We do have that in the docs, but I don't think that should be in the article, it's not intended to explain every detail of how the feature works, just to explain / market the benefits.

yanthomasdev (approved):
LGTM! Good work 🙌

ematipico (approved):

PR #2432, opened 2026-05-28
Migrate to Cloudflare

matthewp:
Migrate astro.build from Netlify to Cloudflare Workers.

__DO NOT MERGE__: There's a few moving parts here to deploy this.


src/app.ts
```diff
@@ -0,0 +1,16 @@
+import { FetchState, astro } from 'astro/fetch';
+
+export default {
+	async fetch(request: Request): Promise<Response> {
+		const url = new URL(request.url);
+
+		// Proxy /fonts/* to fonts-cdn.astro.build
+		if (url.pathname.startsWith('/fonts/')) {
```
florian-lefebvre: I wonder if it would be better to use the cloudflare custom entrypoint to skip astro entirely for rewrites

matthewp: This is only like 2 function calls inside of Astro, it's not really any different.

src/content/blog/welcome-world.mdx
```diff
@@ -20,7 +20,7 @@ import blogImageSeven from "/src/content/blog/_images/welcome-world/7.png"
 For just over three years, Astro has been on a mission to empower the web development community to build a better web. We have undergone plenty of changes over the years, most recently with the release of [Astro 2.0](/blog/astro-2). Today, we are proud to unveil a new look and feel for Astro!
 
 <video autoplay controls loop muted playsinline>
-	<source src="/assets/blog/launch-video.mp4" type="video/mp4" />
+	<source src="https://site-assets.astro.build/blog/launch-video.mp4" type="video/mp4" />
```
matthewp, line 23: This video exceeds the size allowed in Workers, so I pushed it to an R2 bucket and make this domain an alias to it.

src/content.config.ts
```diff
@@ -2,22 +2,7 @@ import { defineCollection } from 'astro:content';
...
-	['css+ui', 'CSS + UI'],
-	['performance+seo', 'Performance + SEO'],
-	['analytics', 'Analytics'],
-	['accessibility', 'Accessibility'],
-	['media', 'Images + Media'],
-	['toolbar', 'Dev Toolbar'],
-	['utilities', 'Utilities'],
-	['uncategorized', 'Uncategorized'],
-] as const);
+import { IntegrationCategories } from './data/integration-categories.js';
```
matthewp, line 5: This prevents the content config from being imported into prod (where it breaks in workerd).

public/_redirects
```diff
@@ -1,6 +1,3 @@
-# Rewrites
-/fonts/*                https://fonts-cdn.astro.build/:splat 200
```
matthewp, line 2: This domain return a 403 and as far as I can tell is not actually used.

pnpm-workspace.yaml
```diff
@@ -8,6 +8,7 @@ allowBuilds:
   esbuild: false
   puppeteer: false
   sharp: false
+  workerd: true
```
delucis: Should this be `false` or does workerd require unsafe postinstall script usage?
```suggestion
  workerd: false
```

README.md
```diff
@@ -27,3 +27,27 @@ Integration data is updated weekly by a [GitHub Action](/.github/workflows/weekl
...
+   ```
+
+2. Upload the file to the `site-assets` R2 bucket:
+   ```sh
+   wrangler r2 object put site-assets/<path> --file <local-file> --content-type <mime-type>
+   ```
+   For example:
+   ```sh
+   wrangler r2 object put site-assets/blog/launch-video.mp4 --file ./launch-video.mp4 --content-type video/mp4
+   ```
```
delucis: Is there an auth step missing here? I assume not just everyone is able to put stuff in this bucket?

public/.assetsignore
```diff
@@ -0,0 +1,2 @@
+_worker.js
+_routes.json
```
delucis: Why was this needed? Would it be something impacting other Cloudflare users that we should upstream to the adapter?

public/_redirects
delucis: It is used for all fonts on the site — `fonts-cdn.astro.build` is served from https://github.com/withastro/fonts-cdn and is gated on `Origin` header to only allow usage from our sites, which is why you saw a 403 when directly accessing it.

The rewrite here instead of using the subdomain directly allows us to avoid an additional server connection and speed up font loading.

I’d assume that’s why fonts are broken in the preview deployment: 
image

delucis (commented):
Left some quick notes on the changes I spotted.

As mentioned in our Discord chat:

- We should make `/agencies` and `/themes` static given Cloudflare does not support CDN cache in front of workers currently (LMK if you’d like me to tackle that)
- I can’t actually see the deploy previews currently, getting an error page:
  Error 1050 page

README.md
matthewp: Yep, I added a comment about this.

public/.assetsignore
matthewp: Not needed, this is autogenerated, so it was removed.

pnpm-workspace.yaml
matthewp: Not needed and removed.

matthewp:
https://github.com/withastro/astro.build/pull/2432/changes/251286d1fe3ca2bd24cbb66189b6b8a07897247e#diff-841254fe75488c1bd4cd7f68f00b4be0e48dcfbc4a16b45847b68295e0e3b27bR8-R15

Moved the font proxying to inside of the worker. I see similar load times for the fonts.

src/pages/agencies/index.astro
```diff
@@ -8,7 +8,7 @@ import GemIcon from '~/components/GemIcon.astro';
 import MainLayout from '~/layouts/MainLayout.astro';
 import BecomeAPartnerCTA from './_components/BecomeAPartnerCTA.astro';
 
-export const prerender = false;
+export const prerender = true;
 Astro.response.headers.set('CDN-Cache-Control', 'public, max-age=60, stale-while-revalidate=86400');
```
delucis: I’m a bit surprised setting headers doesn’t throw an error — I thought it did on static pages — but we should probably still comment this out and leave a reminder to bring it back when possible:

```suggestion
// TODO: re-enable once Cloudflare supports CDN-Cache-Control in front of a worker
// export const prerender = false;
// Astro.response.headers.set('CDN-Cache-Control', 'public, max-age=60, stale-while-revalidate=86400');
```

src/pages/themes/index.astro
delucis, line 19: I’m realising the change to a static page will break the redirect.

Discussing on Discord whether we need to keep it or not: https://discord.com/channels/830184174198718474/1473314557593387100/1514206476212895795

astro.config.mjs
```diff
@@ -1,20 +1,20 @@
...
-import { defineConfig } from 'astro/config';
+import { defineConfig, sessionDrivers } from 'astro/config';
 import astroExpressiveCode from 'astro-expressive-code';
 import icon from 'astro-icon';
 import houston from './houston.theme.json';
 
-/* https://docs.netlify.com/configure-builds/environment-variables/#read-only-variables */
-const NETLIFY_PREVIEW_SITE = process.env.CONTEXT !== 'production' && process.env.DEPLOY_PRIME_URL;
+/* On Cloudflare Workers Builds, CF_PAGES_URL is set to the preview URL for non-production deploys */
+const PREVIEW_SITE = process.env.CF_PAGES_URL;
```
delucis: Are you sure this is correct?

In the preview deploy I’m seeing everything using `astro.build` and not a preview-specific URL, e.g. og:image URLs etc.

matthewp: Updated to use WORKERS_CI_BRANCH which seems to fix it.

src/pages/themes/index.astro
matthewp: Moved this back to SSR then: 4aafcb84b37992a515da204c7f7b2660a3d4548c

astro.config.mjs
delucis: Looks good — we should check after merging to make sure this is also the value we want in prod (i.e. it should be `https://cloudflare-migration.previews.astro.build/` or whatever for previews, but `https://astro.build/` for prod).

src/pages/themes/index.astro
delucis: OK — we can monitor performance. Hopefully it’s fine and Cloudflare will catch up to Netlify’s capabilities soon.

public/_redirects
delucis, line 61: I guess none of these that point to an external domain work anymore?

I tested https://cloudflare-migration.previews.astro.build/chat which we use a lot, but we should handle all of these ideally.

delucis (changes requested):
Looks like our patch which blocks animated images in the image service (`patches/astro.patch`) is no longer working, compare the first image on 

- https://astro.build/themes/details/fortify-astro/
- https://cloudflare-migration.previews.astro.build/themes/details/fortify-astro/

It’s to the sharp image service, maybe Cloudflare uses a different one?

Also spotted something around redirects.

Other than that, there’s a lockfile conflict which will need resolving before merging.

public/_redirects
matthewp: https://cloudflare-migration.previews.astro.build/showcase/submit/ does work. Not sure why the chat one does not. Will figure it out and check all of them.

matthewp: Actually https://cloudflare-migration.previews.astro.build/chat/ (trailingSlash) does work. So something with the config.

delucis:
Ah, re: the image service, @TheOtterlord pointed out that it could be because those images are on on-demand pages and so there’s no optimisation enabled at all by default on CF? On Netlify we do optimise on-demand images.

public/_redirects
matthewp: Oh yeah, it's not config related, it just doesn't automatically redirect both teh trailing slash and non-trailing slash version.

matthewp:
Ok, resolved most everything here:

1. Redirects to non-trailing slash weren't working because in Cloudflare the file expects exact matches. Fixed by adding duplicates.
2. Switched to `cloudflare-binding` to get on-demand optimizations as well. This doesn't fix the animation issue. I think it's supported by the image binding in Cloudflare, but I'll need to add an option to the adapter to allow that to be configured. Will have that in a future PR.

Removed the patch as well. We really shouldn't be patching Astro internals as workarounds. If Astro lacks a needed feature we should add it, and feel the pain like all of our users until it is.

Ready for another review.

matthewp:
Or wait, maybe not, build issue it seems.

matthewp:
Ok, ready now.

public/_redirects
delucis: Oh wow, thanks Cloudflare 😆 

OK, well, verbose, but it works! Good job figuring it out.

astro.config.mjs
```diff
@@ -56,8 +58,28 @@ export default defineConfig({
...
+				// with non-Node.js compatible environments like Cloudflare.
+				'astro-expressive-code/components',
+				'astro-expressive-code>hast-util-select',
+				'astro-expressive-code>rehype',
+				'astro-expressive-code>unist-util-visit',
+				'astro-expressive-code>rehype-format',
+				'astro-expressive-code>hastscript',
+				'astro-expressive-code>hast-util-from-html',
+				'astro-expressive-code>hast-util-to-string',
+				'astro-expressive-code>@expressive-code/core>postcss',
```
delucis, line 76: Not sure if it’s related to this, but the landing page render seems broken:

Image

Looks like the page rendered halfway and didn’t render the rest?

There’s an Expressive Code `<Code>` component exactly there on the page, so I guess maybe related? MDX code blocks in blog posts seem to render fine though.

delucis (changes requested):
Looking good mostly!

Agree re: patching — it’s just an unusual need (“disallow animated images”) so it was hard to prioritise in Astro itself and this was a quick hack for what is fundamentally an issue in the theme portal (which we intentionally have not been investing time in). So yeah, limited resources => suboptimal solutions. But an image component option that disables animation could be handy for sure (`<Image {src} alt="" noAnimation />` or whatever) — not sure if it would be supported in all target image services though.

astro.config.mjs
matthewp: This is fixed now. Just had 1 missing optimizeDeps.

astro.config.mjs
```diff
@@ -1,27 +1,30 @@
...
 import astroExpressiveCode from 'astro-expressive-code';
 import icon from 'astro-icon';
 import houston from './houston.theme.json';
 
-/* https://docs.netlify.com/configure-builds/environment-variables/#read-only-variables */
-const NETLIFY_PREVIEW_SITE = process.env.CONTEXT !== 'production' && process.env.DEPLOY_PRIME_URL;
+/* On Cloudflare Workers Builds, WORKERS_CI_BRANCH is set to the branch name for non-production deploys */
+const PREVIEW_SITE = process.env.WORKERS_CI_BRANCH
+	? `https://${process.env.WORKERS_CI_BRANCH}.previews.astro.build`
+	: undefined;
```
delucis, line 15: Reminder for us to double check this once the PR merges.

astro.config.mjs
```diff
@@ -56,8 +59,29 @@ export default defineConfig({
...
+				'astro-expressive-code>hast-util-from-html',
+				'astro-expressive-code>hast-util-to-string',
+				'astro-expressive-code>@expressive-code/core>postcss',
+				'astro-expressive-code>@expressive-code/core>postcss-nested',
+			],
+		},
 	},
-	adapter: netlify({ imageCDN: false }),
+	adapter: cloudflare({
+		imageService: 'cloudflare-binding',
```
delucis, line 82: Noting that images feel noticeably slower in the preview deploy. I think this is because Netlify had a kind of hybrid set-up:

- images on static pages are optimized at build time and served from the CDN
- images on dynamic pages are optimized on demand

But with `cloudflare-binding` all images are served on-demand.

Not a blocker necessarily, but worth looking into improving. I know @OliverSpeir had one suggestion in https://github.com/withastro/astro/pull/15662 and there’s also https://github.com/withastro/astro/pull/16194 which is related.

delucis (approved):
Congrats @matthewp I believe you have slain the final boss!

Cloudflare sure doesn’t make things easy 😅 

Left two final notes for us to bear in mind post-merge.

I’d also love to figure out the PR comments situation as the stuff CF posts is not amazing, but we can see what it looks like in real-world PRs first.

matthewp:
Merging, will check all of these things after DNS is updated.

PR #2477, opened 2026-06-18
Upgrade to Astro v7

matthewp:
- astro 6.4.4 → 7.0.0
- @astrojs/mdx 6.0.2 → 7.0.0
- @astrojs/netlify 7.0.12 → 8.0.0
- Remove `experimental.rustCompiler` (now default)
- Add `@astrojs/markdown-remark` with `unified()` processor (MDX files use `await` in JSX expressions)
- Temporarily relax `minimumReleaseAge` for beta packages


astro.config.mjs
```diff
@@ -57,9 +58,11 @@ export default defineConfig({
 			noExternal: ['smartypants'],
 		},
 	},
+	markdown: {
+		processor: unified(),
```
matthewp: I didn't know about saterri compat and didn't want to deal with it in this PR.

ArmandPhilippot: Looking at the Astro config file, it seems we do not use any remark/rehype plugin, so Sätteri should work (ie. we don't need to port anything but, yeah, we might want to double-check existing blog posts).

package.json
```diff
@@ -20,19 +20,20 @@
 	},
 	"dependencies": {
 		"@astro-community/astro-embed-youtube": "^0.5.10",
-		"@astrojs/check": "0.9.9",
-		"@astrojs/mdx": "6.0.2",
-		"@astrojs/netlify": "7.0.12",
-		"@astrojs/rss": "4.0.18",
-		"@astrojs/sitemap": "3.7.3",
-		"@astrojs/compiler-rs": "^0.2.1",
+		"@astrojs/check": "0.9.7-beta.1",
```
ArmandPhilippot: Not really related to this PR (this is more about Astro 7) but, I noticed the previous version was `"0.9.9`. This looks like a downgrade. Won't this cause problems at release time?

Same for `@astrojs/rss` and `@astrojs/sitemap`.

I thought https://github.com/withastro/astro/pull/17027 would fix that, but it seems https://github.com/withastro/astro/issues/17024 is still valid.

astro.config.mjs
```diff
@@ -57,9 +58,11 @@ export default defineConfig({
 			noExternal: ['smartypants'],
```
ArmandPhilippot: I wonder if we still need this with Sätteri?

pnpm-workspace.yaml
```diff
@@ -10,4 +10,5 @@ allowBuilds:
   sharp: false
 
 # Require dependencies to be at least 3 days old before allowing them to be installed.
-minimumReleaseAge: 4320
+# TODO: restore to 4320 once Astro v7 beta packages age past 3 days
+minimumReleaseAge: 0
```
ArmandPhilippot: Should we use [`minimumReleaseAgeExclude`](https://pnpm.io/settings#minimumreleaseageexclude) instead? It seems more safe to exclude Astro packages rather than allowing everything.

ArmandPhilippot (commented):
I haven't checked the preview yet, so this is more an initial review. One of my comment is more a concern about Astro 7 release than migrating to Astro 7; you might know better how this works.

astro.config.mjs
matthewp: Might as well dogfood I guess, will update.

trueberryless (approved):
Overall looks good to me 7️⃣ 

I just found a new warning when building locally which is not logged on `main`:

Image

Removing the invalid `:[&:nth-child]` part in the class of the links of the sponsors [here](https://github.com/withastro/astro.build/blob/496d9e240f54c2ea59d0ef8a086615981a77b488/src/pages/_components/landing-page/Sponsors.astro#L173) and [here](https://github.com/withastro/astro.build/blob/496d9e240f54c2ea59d0ef8a086615981a77b488/src/pages/_components/landing-page/Sponsors.astro#L191) removes the warning (notice that no argument is passed to `nth-child`, which is why it is invalid). I think this pseudo selector was unintentionally copied from the other `lg:` breakpoints. All borders are correctly applied even without the pseudo selector.

delucis (approved):
Made a couple of tweaks and browsed various pages — everything looks good to me!

Good catch on that selector warning @trueberryless! I think we can fix that in a follow-up PR as it’s not a new issue, just a newly visible warning.

Also noting that curiously the preview deploys are a little slower than `main` is currently (~1m50s vs ~1m30s comparing last 3 deploys for each). Haven’t investigated why yet — could just be a blip.

PR #2483, opened 2026-06-22
Fix table cell vertical alignment in prose

matthewp:
Changes `align-baseline` to `align-top` on table cells in prose styles. This fixes a visual issue with tables on mobile where baseline alignment was causing layout problems.


ArmandPhilippot (approved):
Seems better to me.

||Before|After|
|---|---|---|
|v7|Astro 7 blog post before|Astro 7 blog post before|
|v4.12|Astro 4.12 blog post before|Astro 4.12 blog post after|

And this one also looks less broken on my side (Cloudflare was a bit truncated):

Image

PR #2485, opened 2026-06-22
Simplify Sätteri comparison table for mobile

matthewp:
Removes backtick formatting from plugin names and inline code examples from the Feature column in the unified/Sätteri comparison table. This reduces the table width on mobile so it doesn't squeeze as badly.


ArmandPhilippot (approved):
It's a bit sad we can't keep inline code for packages name, but this definitely looks better on my side:

Image

PR #2509, opened 2026-07-01
refactor(seo): simplify metadata generation and optimize head ordering

cevdetta:
## Changes

* Simplified the SEO metadata model by making the page metadata (`title`, `description`, `image`, `canonicalURL`, `locale`) the single source of truth.

* Removed duplicated metadata from the `OpenGraph` and `Twitter` types. `OpenGraph` now only exposes Open Graph-specific configuration (`type`), while `Twitter` only exposes Twitter-specific configuration (`card` and `handle`).

* Removed generation of the following Twitter tags, which are already supported through Open Graph fallbacks:

  * `twitter:title`
  * `twitter:description`
  * `twitter:image`
  * `twitter:image:alt`

* Kept the Twitter-specific tags that cannot be inferred from Open Graph:

  * `twitter:card`
  * `twitter:site`

* Removed the duplicated `canonicalURL` property from `OpenGraph`. The page-level `canonicalURL` is now used directly for both:

  * `<link rel="canonical">`
  * `<meta property="og:url">`

* Fixed the canonical URL formatter to correctly normalize trailing slashes.

* Reordered metadata in `BaseHead.astro` to better follow the ordering recommended by Capo.js for `<head>` elements, improving consistency with modern head optimization guidance. ([[Rviscomi](https://rviscomi.github.io/capo.js/)][1])

* Reordered the favicon component so the generated `<head>` better aligns with the expected ordering after all head elements are assembled. ([[Rviscomi](https://rviscomi.github.io/capo.js/)][1])

## Testing

* Verified the generated HTML no longer emits the removed Twitter metadata.

* Confirmed Open Graph metadata continues to render correctly.

* Verified `twitter:card` and `twitter:site` continue to be emitted as expected.

* Verified both `<link rel="canonical">` and `og:url` use the same canonical URL.

* Ran the project locally and confirmed there are no TypeScript or build regressions.

## Resources

Twitter Cards support using several Open Graph properties as fallbacks, making duplicated Twitter metadata unnecessary in most cases.

Although the current documentation has been removed from the X Developer site, it is available through the Internet Archive:

* Twitter Cards Markup (archived):
  [https://web.archive.org/web/20240613190853/https://developer.x.com/en/docs/twitter-for-websites/cards/overview/markup](https://web.archive.org/web/20240613190853/https://developer.x.com/en/docs/twitter-for-websites/cards/overview/markup)

* Getting Started with Twitter Cards (archived):
  [https://web.archive.org/web/20240616101429/https://developer.x.com/en/docs/twitter-for-websites/cards/guides/getting-started](https://web.archive.org/web/20240616101429/https://developer.x.com/en/docs/twitter-for-websites/cards/guides/getting-started)

The documentation notes:

> If an `og:type`, `og:title`, and `og:description` exist in the markup but `twitter:card` is absent, then a Summary Card may be rendered.

Since Astro intentionally emits `summary_large_image`, only `twitter:card` (and optionally `twitter:site`/`twitter:creator`) need to be generated separately, while the remaining metadata can be provided via Open Graph.

The `<head>` reordering is based on the recommendations from Capo.js, which validates and recommends an optimal ordering of head elements to improve browser parsing and perceived performance. ([[Rviscomi](https://rviscomi.github.io/capo.js/)][1])

[1]: https://rviscomi.github.io/capo.js/ "capo.js: get your ﹤𝚑𝚎𝚊𝚍﹥ in order | capo.js"

### DIFF
--- a/src/components/SEO.astro
+++ b/src/components/SEO.astro
@@ -11,10 +11,10 @@ export type SEOMetadata = {
 	canonicalURL?: URL | string | null;
 	locale?: string;
 };
-export type OpenGraph = Partial<SEOMetadata> & {
+export type OpenGraph = {
 	type?: string;
 };
-export type Twitter = Partial<SEOMetadata> & {
+export type Twitter = {
 	handle?: string;
 	card?: 'summary' | 'summary_large_image';
 };
@@ -29,42 +29,29 @@ const {
 	image,
 	locale = 'en',
 	canonicalURL = new URL(Astro.url.pathname, Astro.site),
+	og: ogProps = {},
+	twitter: twitterProps = {},
 } = Astro.props;
 
 const og = {
-	name,
-	title,
-	description,
-	canonicalURL,
-	image,
-	locale,
 	type: 'website',
-	...Astro.props.og,
+	...ogProps,
 } satisfies OpenGraph;
 
 const twitter = {
-	name,
-	title,
-	description,
-	canonicalURL,
-	image,
-	locale,
 	card: 'summary_large_image',
-	...Astro.props.twitter,
+	...twitterProps,
 } satisfies Twitter;
 
 /**
  * Enforce some standard canonical URL formatting across the site.
  */
 function formatCanonicalURL(url: string | URL) {
-	const path = url.toString();
-	const hasQueryParams = path.includes('?');
-	// If there are query params, make sure the URL has no trailing slash
-	if (hasQueryParams) {
-		path.replace(/\/?$/, '');
-	}
-	// otherwise, canonical URL always has a trailing slash
-	return path.replace(/\/?$/, hasQueryParams ? '' : '/');
+  const value = url.toString();
+
+  return value.includes('?')
+    ? value.replace(/\/?$/, '')
+    : value.replace(/\/?$/, '/');
 }
 ---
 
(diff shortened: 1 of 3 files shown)


src/components/SEO.astro
```diff
@@ -11,10 +11,10 @@ export type SEOMetadata = {
 	canonicalURL?: URL | string | null;
 	locale?: string;
 };
-export type OpenGraph = Partial<SEOMetadata> & {
+export type OpenGraph = {
 	type?: string;
 };
-export type Twitter = Partial<SEOMetadata> & {
+export type Twitter = {
```
delucis, line 17: Why were these types removed?

cevdetta: Because neither **Twitter** nor **OpenGraph** are going to use the `SEOMetadata`'s types, they'll directly infer it from `Astro.props`, instead of re-routing them again via `Astro.props.{og,twitter}`.

delucis (commented):
Thanks for the answer @cevdetta.

I had a quick look at this PR and I don’t think it improves anything important. It would take me some time to review it properly given it’s a decent amount of changes, which were submitted without first asking whether we wanted these changes in the first place.

Given we don’t really get any benefit, we won’t merge this at this time.

PR #2520, opened 2026-07-06
chore: remove private beta label from workers cache in v7 post

ghostdevv:
- https://github.com/withastro/docs/pull/14198
- https://blog.cloudflare.com/workers-cache/

### DIFF
--- a/src/content/blog/astro-7.mdx
+++ b/src/content/blog/astro-7.mdx
@@ -403,7 +403,7 @@ Each adapter exports a provider from its `/cache` entrypoint:
 | -- | -- | -- |
 | Netlify | `@astrojs/netlify/cache` | `cacheNetlify()` |
 | Vercel | `@astrojs/vercel/cache` | `cacheVercel()` |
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks @ghostdevv! In general we don’t go back and update blog posts, instead leaving them as “historical” points in time, but as this is a fairly recent one I think this update makes sense 👍

src/content/blog/astro-7.mdx
delucis: Oh, actually, we also need to remove this!
```suggestion
| Cloudflare | `@astrojs/cloudflare/cache` | `cacheCloudflare()` |
```

delucis (changes requested):

ghostdevv:
> Thanks @ghostdevv! In general we don’t go back and update blog posts, instead leaving them as “historical” points in time, but as this is a fairly recent one I think this update makes sense 👍

for sure! definitely agree it's worth for this one though, I imagine people will reference this when upgrading Astro (I always do)

delucis (approved):
Let’s go! Thanks again @ghostdevv 💜

PR #2539, opened 2026-07-14
Fix 404 link to Astro development case study

arlindmusliu:
Updated the link to the Astro development case study.

## Browser Test Checklist

- [x] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


delucis (approved):
Thanks for the fix @arlindmusliu!

PR #2490, opened 2026-06-23
INTEGRATION icons: astro-gtm-lite, astro-bootstrap-icons and @dropinb…

pascal-brand38:
…log/astro

Adding icon for integration astro-gtm-lite, astro-bootstrap-icons and @dropinblog/astro

Results checked with
npm run update:integrations -- --unsafe

Closes #2484
Closes #2486
Closes #2487

## Browser Test Checklist

I have tested this PR on at least three of the following browsers:

- [X] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari

PR #2540, opened 2026-07-14
Add astro-chat-widget.svg for integrations

eeegoloauq:
Added SVG Avatar for the `astro-chat-widget` package

## Browser Test Checklist

- [x] Chrome / Chromium
- [ ] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


delucis (approved):
Thanks for the PR @eeegoloauq!

Just adding the SVG isn’t sufficient, it also needs to be referenced in `scripts/integrations.json`. You can see this recent PR for reference: https://github.com/withastro/astro.build/pull/2490

delucis (changes requested):
Thanks for the PR @eeegoloauq!

Just adding the SVG isn’t sufficient, it also needs to be referenced in `scripts/integrations.json`. You can see this recent PR for reference: https://github.com/withastro/astro.build/pull/2490

delucis (approved):
Thanks!

PR #2545, opened 2026-07-15
Add Prime Stack to the showcase

eyal-meshulam:
Adds Prime Stack (https://0587178899.com), an Israeli studio building fast, accessible, SEO-first websites on Astro. Its own site ships a 100 mobile Lighthouse score with self-hosted variable fonts, zero client-side framework JS, GA4 via Cloudflare Zaraz, and a built-in AI chat rep on a Cloudflare Worker.


delucis:
Thanks for the PR @eyal-meshulam! We add new sites through an automated process: you can post a link to your site in this discussion and it will automatically be picked up next Monday: https://github.com/withastro/roadmap/discussions/521

PR #2587, opened 2026-08-02
Update prism integration's url

koeliga:
old link was dead link, this link should link to the correct prism page from the documentation


delucis:
Thanks for the PR @koeliga! However, this data file is autogenerated from the npm registry so this link will need fixing upstream.

The line to change is here if you’d like to make a PR: https://github.com/withastro/astro/blob/0a79753610ff4960ae6c9e6e334115f6ea0bc954/packages/astro-prism/package.json#L14

PR #2598, opened 2026-08-05
Update aspire.dev showcase screenshot

delucis:
This PR updates the screenshot of aspire.dev in the showcase as requested in https://github.com/withastro/roadmap/discussions/521#discussioncomment-17907698


ArmandPhilippot (approved):
They asked for a "dark theme-if that's possible", but since I don't know how we handle screenshots, it might not be an oversight! Thanks, LGTM!

delucis:
> They asked for a "dark theme-if that's possible"

Oh, very good catch, I totally skimmed over that detail 😅 Our normal automated process defaults to light mode, but in this case I’m taking the screenshot manually myself, so it’s feasible to use dark mode. I updated the screenshot.

Thanks for reviewing 💜

PR #2608, opened 2026-08-10
fix(showcase): update Room TBA to its current domain

smmariquit:
Room TBA moved from `room-tba.uplbtools.me` to `room-tba.uplb.tools` after the project moved to the uplbtools org. The old domain still 308s to the new one, so the showcase link works, but it displays the outdated URL.

Renames the entry and its screenshot to match the current domain and updates the `url` field. `dateAdded` is unchanged.

Site: https://room-tba.uplb.tools
Repo: https://github.com/uplbtools/room-tba


delucis (approved):
Thanks for the update @smmariquit

PR #2688, opened 2026-09-07
Add updated Bits&Letters agency page

ddemaree:
Replace old Bits&Letters agency copy with new text that's longer, better, more up-to-date, clearer, and did I mention it's better?

Branch naming notwithstanding, all text was human-written (because Astra's version was not good).

## Browser Test Checklist

- [X] Chrome / Chromium
- [X] Firefox
- [ ] Android Firefox
- [X] Safari
- [ ] iOS Safari

### DIFF
--- a/src/content/agencies/bits-and-letters.md
+++ b/src/content/agencies/bits-and-letters.md
@@ -25,19 +25,13 @@ Check out our website for more about [our services](https://www.bitsandletters.c
 
 ### What clients say about us
 
-<blockquote>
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks @ddemaree! I made one small tweak to make the subheadings `<h3>`s instead of `<h2>`s so they’re semantically nested under the “About Bits&Letters” heading.

Apologies that there’s no deploy preview on this PR. We don’t get many astro.build PRs from forks, so I hadn’t noticed that’s not enabled for forks since we migrated the site to Cloudflare.

In any case, I checked it locally and it looks great. Welcome back as an agency partner and thanks again for your support! 💜

PR #2691, opened 2026-09-08
Evil martians case study

sarah11918:
draft for previewing

## Browser Test Checklist

- [x] Chrome / Chromium
- [x] Firefox
- [ ] Android Firefox
- [ ] Safari
- [ ] iOS Safari


src/content/caseStudies/evil-martians.mdx
```diff
@@ -0,0 +1,108 @@
...
+		intact, an LLM could handle most of the repetitive translation.
+	</Fragment>
+	<Fragment slot="caption">
+		— Rita Klubochkina, Sr. Frontend Engineer at Evil Martians
+	</Fragment>
+</Blockquote>
+
+Astro's content collections replaced the GraphQL layer, with no changes needed to the team's custom content pipeline at all. The existing LiquidJS and JSX transformers continued to run, just behind content collection loaders instead of Gatsby GraphQL resolvers.
+
+Astro's routing system allowed the migration to progress incrementally, with both frameworks built from the same project repository simultaneously. The team continuously deployed the new Astro project, while the Gatsby project kept serving production. When the migration was complete, Evil Martians switched over to the Astro site with zero risk or downtime. This allowed the migration to progress at its own pace, without affecting production.
```
HiDeoo: > When the migration was complete, Evil Martians switched over to the Astro site with zero risk or downtime.

Isn't this in contradiction with one of the point in the previous sections?

> swapping new routes for old ones as they became available

src/content/caseStudies/evil-martians.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+## Results
+
+Evil Martians was facing an aging framework, a GraphQL layer that was overkill for a static site, and a shrinking pool of engineers to maintain it.
+
+The migration to Astro allowed the team to ship a set of islands, first as full-pages, then with each section hydrating independently. This change alone reduced the amount of JavaScript shipped on load by 61%.
```
HiDeoo: Not sure if it's me, but I find this sentence a bit difficult to parse. It first says a set of islands is shipped, then followed by "as full-page" (I wonder if full-page is the best word here) which can sounds as opposite?

I don't have a suggestion to make yet but will think about it more.

src/content/caseStudies/evil-martians.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+Islands also made it obvious which parts of the site were heavy. The team discovered that the heaviest element on the homepage was the site search modal. Previously, this was loaded up front on every page, whether opened or not. Deferring this to first open as an island removed 437 KB of load-time JavaScript across the site.
+
+Deleting the GraphQL layer also deleted its expensive prefetch behavior. Gatsby pulled 13.6 MB of query results for links nobody had clicked yet, more than six times the weight of the page itself. The Astro version of the site waits for a hover before fetching any data.
```
HiDeoo:
```suggestion
Deleting the GraphQL layer also deleted its expensive prefetch behavior. Gatsby pulled 13.6 MB of query results for links nobody had clicked yet, more than six times the weight of the page itself. The Astro version prefetches linked pages on hover.
```

Could be me but "waiting for a hover" and "any data" sounds a bit weird, maybe we could say we only prefetch a linked page rather than any data and only on hover?

src/content/caseStudies/evil-martians.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+Additionally, Evil Martians improved performance and security by reducing dependencies. 17 Gatsby plugins and 4 patched dependencies are now gone. `package.json` is down from 108 to 86 direct dependencies.
+
+### Performance gains
+
+The migration to Astro was a clear success. The team was able to migrate the entire site in just eight days, without rewriting a single React component. The homepage went from 2.1 MB to 1.8 MB initially, and after fully converting to islands architecture, it's now 1.1 MB. Homepage JavaScript at load fell 61%, and mobile Lighthouse performance rose from 66 to 90.
```
HiDeoo: Maybe on purpose but this part could feel a bit repetitive I guess? Could it be just something like "The following table breaks down the various savings in homepage asset sizes." or something similar?

src/content/caseStudies/evil-martians.mdx
delucis: I’ll be honest that this section was a bit unclear to me so I edited it to describe it like this, but I wasn’t clear from our notes what their exact process was. It seems clear they migrated route by route, but it wasn’t clear to me whether that meant each route got deployed to production as they migrated or whether they only switched to production once all routes were migrated.

src/content/caseStudies/evil-martians.mdx
delucis: Maybe something along these lines?

```suggestion
The migration to Astro allowed the team to progressively refactor existing React components, eventually using Astro’s fine-grained control to hydrate each section independently. This change alone reduced the amount of JavaScript shipped on load by 61%.
```

src/content/caseStudies/evil-martians.mdx
delucis: I think the “waits” is semi-important because it’s trying to underline that the Astro behaviour is less aggressive. But I can see what you mean.

This is maybe more verbose but could be clearer?

```suggestion
Deleting the GraphQL layer also deleted its expensive prefetch behavior. Gatsby pulled 13.6 MB of query results for links nobody had clicked yet, more than six times the weight of the page itself. Astro’s prefetch feature optimizes navigations, but is much less aggressive, only prefetching linked pages on hover.
```

src/content/caseStudies/evil-martians.mdx
HiDeoo: This does indeed sounds a lot better imo and way easier to read.

src/content/caseStudies/evil-martians.mdx
sarah11918: I believe the intention was that building the new site happened route (group) by route group, and it built alongside the full Gatsby site, but that the "flicking of the switch" happened all at once. In an earlier version they did mention that this process allowed them to "pause when/if they needed to" and nothing would affect production.

Maybe the idea of "doing some routes first" isn't that big of a deal to mention publicly, but I think in aggregate the idea was, it only took wrapping things as islands, bit by bit, so it could all happen pretty quickly.

src/content/caseStudies/evil-martians.mdx
sarah11918:
```suggestion
The migration to Astro was a clear success. The following table breaks down the various savings in homepage asset sizes.
```

src/content/caseStudies/evil-martians.mdx
sarah11918:
```suggestion
```
We probably don't lose anything if we elminate this!

theotterlord (approved):

